PR #6, opened 2016-02-19
Add {{group-by}} helper

martndemus:


martndemus:
Should we drop support for 1.13?

README.md
```diff
@@ -26,6 +26,23 @@ Pipes the return values of actions in a sequence of actions.
...
 ```
 
+### Group By
+
+Returns an object where the keys are the unique values of the given property,
+the values are an array with all items of the array that have the same value of
+that property.
+
+```hbs
+{{#each-in (group-by artists 'category') as |category artists|}}
```
poteto: Double quotes please! :)

addon/helpers/group-by.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default Helper.extend({
+  compute([array, byPath]) {
+    set(this, 'array', array);
+    set(this, 'byPath', byPath);
+
+    return get(this, 'content');
+  },
+
+  array:   null,
```
poteto: Small nitpick but prefer not to align values like this

addon/helpers/group-by.js
```diff
@@ -0,0 +1,59 @@
...
+  byPath:  null,
+  content: null,
+
+  byPathDidChange: observer('byPath', function() {
+    let byPath = get(this, 'byPath');
+
+    if (byPath) {
+      defineProperty(this, 'content', computed(`array.@each.${byPath}`, groupFunction));
+    } else {
+      defineProperty(this, 'content', null);
```
poteto: Return an empty object? What happens when you use `each-in` on `null`?

tests/integration/helpers/group-by-test.js
```diff
@@ -0,0 +1,50 @@
...
+  ]));
+
+  this.render(hbs`
+    {{~#each-in (group-by array 'category') as |category entries|~}}
+      {{~category~}}
+      {{~#each entries as |entry|~}}{{~entry.name~}}{{~/each~}}
+    {{~/each-in~}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'aabbcd');
```
poteto: Assertion messages please! 🤓

poteto:
Hmm, I was looking for an `each-in` polyfill but couldn't find one. Maybe we should make it and use it as a dependency here?

martndemus:
We could polyfill `each-in` with a component that does `{{yield key value}}` for each kv pair.

app/helpers/group-by.js
```diff
@@ -0,0 +1 @@
+export { default, groupBy } from 'ember-functional-helpers/helpers/group-by';
```
poteto: Can we also export in https://github.com/DockYard/ember-functional-helpers/blob/remove-compose/addon/index.js please

martndemus:
@homu r+

martndemus:
@homu r+

poteto:
Rebase please

martndemus:
@homu r+

addon/helpers/group-by.js
henrymazza: @poteto  Got bitten by that. Helper just fails and something swallows all the errors rendering a blank page. Perhaps best behavior should be discard null groupNames unless a `defaultGroupName` provided, like:

```handlebars
{{each-in (group-by "category" "uncategorized" authors)}}
```

PR #24, opened 2016-02-19
Add Range helper

### LINKED ISSUE #22: Range helper
```js
range(10, 12) === [10, 11];
range(10, 12, false) === [10, 11];
range(10, 12, true) === [10, 11, 12];
range(-1, 2, true) === [-1, 0, 1, 2];
range(4, 1) === [4, 3, 2];
```

```hbs
{{#each (range 10 20 true) as |number, index|}}
  {{index}}: {{number}}
{{/each}}
```

```hbs
{{#each (range min max) as |number, index|}}
  {{index}}: {{number}}
{{/each}}
```

poteto:
Closes #22

### DIFF
--- a/addon/helpers/range.js
+++ b/addon/helpers/range.js
@@ -1,0 +1,37 @@
+import Ember from 'ember';
+import { gte, lte, gt, lt } from '../utils/comparison';
+
+const {
+  Helper: { helper },
+  typeOf
+} = Ember;
+
+export function range([min, max, isInclusive]) {
+  isInclusive = typeOf(isInclusive) === 'boolean' ? isInclusive : false;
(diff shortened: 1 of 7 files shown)


addon/helpers/range.js
```diff
@@ -0,0 +1,37 @@
...
+  let testFn = isInclusive ? lte : lt;
+  let numbers = [];
+
+  if (min < max) {
+    for (let i = min; testFn(i, max); i++) {
+      numbers.push(i);
+    }
+  }
+
+  if (min > max) {
```
poteto: My brain isn't working too well today, I suppose there is probably a better way of doing this

@martndemus

poteto:
@homu r+

PR #40, opened 2016-02-20
Tree shake helpers using `only` or `except`

### LINKED ISSUE #29: Tree shake unused helpers

poteto:
Closes #29

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,4 +41,5 @@ install:
   - bower install
 
 script:
+  - npm run-script node-test
   - ember try $EMBER_TRY_SCENARIO test
--- a/README.md
+++ b/README.md
@@ -17,9 +17,25 @@ To install:
 ember install ember-composable-helpers
 ```
 
-### See also:
+## Configuration
+This addon performs optional tree-shaking – you can specify which helpers to whitelist or blacklist using `only` or `except` within your `config/environment.js`:
 
-* [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers)
+```js
+module.exports = function(environment) {
+  var ENV = {
+    'ember-functional-helpers': {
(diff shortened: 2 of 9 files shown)


poteto:
@martndemus :eyes: please

martndemus:
LGTM, but a few things:
- can you remove the compose artifact in another PR?
- should we add tests for the tree shaking?

martndemus:
One more thing: should we wanr on `test` or `development` that a helper has been filtered out? E.g.:

```js
export default Ember.Helper.helper(() => assert('The ${helperName} helper has been disabled, if you want to use this helper please enable it in your config file'));
```

poteto:
I was actually unsure how to test. If I add the config to the addon's `config/environment` or `dummy/config/environment`, it will actually remove the helper and break tests...

poteto:
If the user tries to use a helper that is excluded it will throw an error. A warning doesn't seem necessary

martndemus:
For tests: We can use `mocha` to test on the Node.js level

For the warning: It might prevent someone losing an hour of his life because he was checking if the addon was installed correctly.

poteto:
Tests have been added.

The excluded helper will not be included at all, so we cannot perform an assertion within the individual helper definition. 

> For the warning: It might prevent someone losing an hour of his life because he was checking if the addon was installed correctly.

I disagree with this. If the user has explicitly opted-in to whitelisting or blacklisting specific helpers, they should already be aware that not all helpers will be available. By default, all helpers are included – these white/blacklists are specifically for optimization.

Further, Ember will already throw a JavaScript error in the console when you try to use a helper that doesn't exist. A warning would be redundant.

martndemus:
Tests look good, let's merge this!

poteto:
@homu r+

Issue #56, opened 2016-02-22
Compact helper


martndemus:
```js
compact([1, 2, null, 3]) === [1, 2, 3];
compact([,,,,,]) === [];
```
```hbs
{{#each (compact arrayWithNulls) as |item|}}

{{/each}}
```

poteto:
Should also work for objects

```js
compact({ foo: 'bar', items: undefined, things: null }) === { foo: 'bar' };
compact({ foo: null }) === {};
```
```hbs
{{#each-in (compact someObj) as |k, v|}}
  {{! stuff }}
{{/each}}
```

martndemus:
:+1:

PR #64, opened 2016-02-22
Add Compact Helper

spencer516:
For: https://github.com/DockYard/ember-composable-helpers/issues/56

I thought I'd take a stab at this one — but, the requirement to compact objects leaves me a bit flummoxed. Per the [`each-in` helper](https://github.com/emberjs/ember.js/blob/v2.3.0/packages/ember-htmlbars/lib/helpers/each-in.js#L8), it's "unbound, in that new (or removed) properties added to the target object will not be rendered". 

So even though this compact helper would be dynamically adding/removing keys, the `each-in` helper would never update to render the new/removed keys. 

This PR does not attempt to address compacting based on object values yet — can add that if, even given the constraints, y'all would like to have that added. :)

### DIFF
--- a/addon/helpers/compact.js
+++ b/addon/helpers/compact.js
@@ -1,0 +1,32 @@
+import Ember from 'ember';
+
+const {
+  Helper,
+  get,
+  set,
+  observer,
+  isPresent,
+  isArray,
+  A: emberArray,
+  computed: { filter }
+} = Ember;
+
+export default Helper.extend({
+  array: null,
+
+  compute([array]) {
+    if (!isArray(array)) {
+      return emberArray([array]);
+    }
+
+    set(this, 'array', array);
+
+    return get(this, 'content');
+  },
+
+  content: filter('array', isPresent),
+
+  contentDidChange: observer('content', function() {
+    this.recompute();
+  })
+});
--- a/tests/integration/helpers/compact-test.js
+++ b/tests/integration/helpers/compact-test.js
@@ -1,0 +1,62 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+
+const { A: emberArray, run } = Ember;
+
+moduleForComponent('compact', 'Integration | Helper | {{compact}}', {
+  integration: true
+});
+
+test('Removes empty values in standard arrays', function(assert) {
+  this.set('array', emberArray([1, 2, null, 3, false]));
+  this.render(hbs`
+    {{~#each (compact array) as |value|~}}
+      {{value}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '123false', 'null is removed');
+});
+
+test('It graceully handles non-array values', function(assert) {
+  this.set('notArray', 1);
+  this.render(hbs`
+    {{~#each (compact notArray) as |value|~}}
+      {{value}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '1', 'the non array value is rendered');
+});
+
+test('It recomputes the filter if the array changes', function(assert) {
+  this.set('array', emberArray([1, 2, null, 3]));
+  this.render(hbs`
+    {{~#each (compact array) as |value|~}}
+      {{value}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '123', 'null is removed');
+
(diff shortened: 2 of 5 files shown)


addon/helpers/compact.js
```diff
@@ -0,0 +1,27 @@
...
+export default Helper.extend({
+  compute([array]) {
+    set(this, 'array', array);
+    return get(this, 'content');
+  },
+
+  array: null,
+
+  content: filter('array', function(value) {
+    return !isEmpty(value);
```
martndemus: you could use `isPresent` here.

tests/integration/helpers/compact-test.js
```diff
@@ -0,0 +1,54 @@
...
+  this.set('array', array);
+  this.render(hbs`
+    {{~#each (compact array) as |value|~}}
+      {{value}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '123', 'null is removed');
+
+  run(() => {
```
martndemus: This can be a one-liner.

martndemus:
Looks good, we are aware of the inability to recompute on object keys. So if you would like, please do add the functionality for objects.

martndemus:
Please also add the helper to: https://github.com/DockYard/ember-composable-helpers/blob/master/addon/index.js

spencer516:
^^ will do (to both)

poteto:
Something like this could handle compacting objects:

```js
export function compact(objectInstance) {
  const compactedObject = {};

  for (let key in objectInstance) {
    const value = objectInstance[key];

    if (Ember.isPresent(value)) {
      compactedObject[key] = value;
    }
  }

  return compactedObject;
}
```

```js
test('#compact returns an object with all `null` and `undefined` elements removed', function(assert) {
  const rawData = {
    firstName: 'Michael',
    lastName: 'Bolton',
    company: undefined,
    age: null,
    favoriteDrink: ''
  };
  const expectedResult = {
    firstName: 'Michael',
    lastName: 'Bolton'
  };

  const result = objectTransforms.compact(rawData);
  assert.deepEqual(result, expectedResult, 'it should not contain `null` or `undefined` elements');
});
```

poteto:
For now I think we can leave this as an array compact, and revisit the object compact helper in a separate PR. Thanks!

tests/integration/helpers/compact-test.js
```diff
@@ -0,0 +1,52 @@
...
+import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+
+const { A: emberArray, run } = Ember;
+
+moduleForComponent('compact', 'Integration | Helper | {{compact}}', {
+  integration: true
+});
+
+// Replace this with your real tests.
```
poteto: Kill this, please!

addon/helpers/compact.js
```diff
@@ -0,0 +1,27 @@
...
+  get,
+  set,
+  observer,
+  isPresent,
+  computed: { filter }
+} = Ember;
+
+export default Helper.extend({
+  compute([array]) {
+    set(this, 'array', array);
```
poteto: We should add a guard here for when `array` is not actually an `array` using `isArray`. In this scenario we can just return the value wrapped in an array:

```js
if (!isArray(array)) {
  return [array];
}
```

martndemus:
I think a small section in `README.md` would also be welcome :)

martndemus:
This one looks good to merge, can you please squash all your commits into a single commit?

spencer516:
Squashed, sir

martndemus:
@homu r+

addon/helpers/compact.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    set(this, 'array', array);
+
+    return get(this, 'content');
+  },
+
+  array: null,
+
+  content: filter('array', function(value) {
```
poteto: This can just be `filter('array', isPresent)

poteto:
@homu r-

addon/helpers/compact.js
```diff
@@ -0,0 +1,34 @@
...
+    if (!isArray(array)) {
+      return emberArray([array]);
+    }
+
+    set(this, 'array', array);
+
+    return get(this, 'content');
+  },
+
+  array: null,
```
poteto: Please move this to the L15, we prefer to group props first, then computed/observers

poteto:
Perfect, thanks! @homu r+

PR #72, opened 2016-02-22
Compact object

spencer516:
This adds support to https://github.com/DockYard/ember-composable-helpers/pull/64 for compacting objects. 

FYI: Tests are failing on Beta/Canary due to this bug: https://github.com/emberjs/ember.js/issues/12995

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ For action helpers, this will mean better currying semantics:
   + [`without`](#without)
 * [Object](#object-helpers)
   + [`group-by`](#group-by)
+  + [`compact`](#compact)
 * [Math](#math-helpers)
   + [`inc`](#inc)
   + [`dec`](#dec)
@@ -426,12 +427,20 @@ Joins the given array with an optional separator into a string.
 **[⬆️ back to top](#available-helpers)**
 
 #### `compact`
-Removes blank items from an array.
+Removes blank items from an array or blank values from an object.
 
 ```hbs
 {{#each (compact arrayWithBlanks) as |notBlank|}}
   {{notBlank}} is most definitely not blank!
 {{/each}}
+```
+
+Or:
+
+```hbs
+{{#each-in (compacy objectWithBlanks) as |notBlankKey notBlankValue|}}
+  {{notBlankValue}} is also definitely not blank.
+{{/each-in}}
 ```
 
 **[⬆️ back to top](#available-helpers)**
--- a/addon/helpers/compact.js
+++ b/addon/helpers/compact.js
@@ -3,25 +3,54 @@ import {
   A as emberArray,
   isEmberArray as isArray
 } from 'ember-array/utils';
-import { filter } from 'ember-computed';
+import computed from 'ember-computed';
 import Helper from 'ember-helper';
 import get from 'ember-metal/get';
 import observer from 'ember-metal/observer';
 import set from 'ember-metal/set';
-import { isPresent } from 'ember-utils';
+import { isPresent, typeOf } from 'ember-utils';
+const { defineProperty } = Ember;
+const { keys } = Object;
+
+// For compacting object instances
+export function compact(objectInstance) {
+  return keys(objectInstance).reduce((compactedObject, key) => {
+    let val = objectInstance[key];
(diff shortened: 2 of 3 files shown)


addon/helpers/compact.js
```diff
@@ -0,0 +1,66 @@
...
+  get,
+  set,
+  observer,
+  isPresent,
+  computed,
+  copy,
+  typeOf,
+  A: emberArray
+} = Ember;
+const keys = Object.keys || Ember.keys;
```
poteto: We can kill `Ember.keys`, that is deprecated.

addon/helpers/compact.js
```diff
@@ -0,0 +1,66 @@
...
+  }, {});
+}
+
+export default Helper.extend({
+  compute([arrayOrObject]) {
+    set(this, 'arrayOrObject', arrayOrObject);
+
+    if (isArray(arrayOrObject)) {
+      set(this, 'keys', []);
+    } else if (typeOf(arrayOrObject) === 'object') {
```
martndemus: also check for `'instance'`

addon/helpers/compact.js
```diff
@@ -0,0 +1,66 @@
...
+      return emberArray([arrayOrObject]);
+    }
+
+    return get(this, 'content');
+  },
+
+  arrayOrObject: null,
+  keys: null,
+
+  keysToWatchDidChange: observer('keys', function() {
```
poteto: `keys.[]`?

martndemus: I don't think we should watch for the object his keys at all. Keys added to the object will not be watched, which will result in all kinds of weird behaviour. Let's keep it simple.

poteto: Then why make it an observer at all? We can perform this logic in the `compute` method if it doesn't need to update.

martndemus: We only need the observer when the watched object is an array.

poteto: Isn't that the observer on L63? This observer seems to be for the object.

poteto: Oh right, I think we're both in agreement that this observer can be removed

martndemus: The observer for L63 is to trigger the recompute when the end result changes. This observer is to update to reflect the input of the helper.

addon/helpers/compact.js
spencer516: It is; but doesn't 1.13 still support IE 8, which needs the Ember.keys polyfill?

spencer516:
^^ Did a bit of refactoring per your recommendations and got rid of that superfluous observer.

spencer516:
I rebased this as it got a bit stagnant. If y'all have any other feedback, let me know.

poteto:
If you rebase this against master, `beta` will be an allowed failure.

addon/helpers/compact.js
```diff
@@ -2,27 +2,60 @@ import Ember from 'ember';
...
   isPresent,
-  isArray,
-  A: emberArray,
-  computed: { filter }
+  computed,
+  copy,
+  typeOf,
+  A: emberArray
 } = Ember;
+const keys = Object.keys || Ember.keys;
```
martndemus: This line is not needed, just use `Object.keys`.

spencer516:
Sorry I hadn't touched this in a while — must have missed that comment from 15 days ago.

taras:
Let's make this happen :)

README.md
```diff
@@ -426,14 +427,22 @@ Joins the given array with an optional separator into a string.
...
 ```hbs
 {{#each (compact arrayWithBlanks) as |notBlank|}}
   {{notBlank}} is most definitely not blank!
 {{/each}}
 ```
 
+Or:
+
+```hbs
+{{#each-in (compacy objectWithBlanks) as |notBlankKey notBlankValue|}}
```
poteto: `s/compacy/compact/`

martndemus:
We're not going to merge this, as compact cannot track object properties being added/removed, so it would lead to weird stuff happening updating the object after the helper has run once.

Issue #73, opened 2016-02-22
Merge Ember-Curry-Helpers


flexyford:
I'm really excited to see Ember moving in the direction of declarative templating with `ember-composable-helpers`. I've been working with composing actions and creating explicit handlebars templates with [ember-curry-helpers](https://github.com/flexyford/ember-curry-helpers). This project was based on [ramdJs](http://ramdajs.com/0.19.1/), a functional programming library. I gave an [Ember Meetup Talk](https://www.youtube.com/watch?v=otIpuV9Q4Eg) in January where I built a TODO application with composable curried actions using `ember-curry-helpers`. 

I'm hoping that `ember-composable-helpers` would benefit from some things I've learned from working on `ember-curry-helpers` and consider merging some functionality into this existing repository.

##### Lessons Learned from Ember-Curry-Helpers
- The `pipe` helper passes the return value of the currently evaluated function, as the argument to the next function in the sequence of actions/functions.

- In fact, all functions would benefit from being curried functions. We can then compose more explicit handlebars templates. In the example below, `append` exists as an Ember-Composable-Array-Helper which can accept arguments that are curried-in. The `addToCart` action does not need to exist on the controller anymore.

```hbs
{{!Pipe example with curried function `append`}}
{{!<button {{action (pipe addToCart purchase redirectToThankYouPage) item}}>}}
<button {{action (pipe (append cart) purchase redirectToThankYouPage) item}}>
  1-Click Buy
</button>
```

- `Pipe` is promise aware. In the above example, we would `redirectToThankYouPage` only if `purchase` returns a resolved Promise.

##### Contributing

Rather than duplicate our efforts​, myself and @cowboyd would like to collaborate and merge in this functionality from `ember-curry-helpers` and continue along with with just `ember-composable-actions`.

Of course we'd  be happy to submit a PR for all these things, so let us know if this fits into your vision of `ember-composable-helpers`.

poteto:
I am +1 on this. @cowboyd's post on functional templating + the work done in `ember-curry-helpers` resonates well with the direction I want to take with this addon. I'm not sure if we want to port `ramda.js` to Ember, but I think a good vision to go by is that we want this addon to provide composable primitives for functional templating.

> The pipe helper passes the return value of the currently evaluated function, as the argument to the next function in the sequence of actions/functions.

This addon's `pipe` helper already handles this. 

[...]

PR #78, opened 2016-02-23
Fix return value passing in `pipe` helper

poteto:
For #73

### DIFF
--- a/addon/helpers/pipe.js
+++ b/addon/helpers/pipe.js
@@ -12,7 +12,7 @@ export function pipe(actions = []) {
         return curr(...args);
       }
 
-      return curr(acc, ...args);
+      return curr(acc);
     }, undefined);
   };
 }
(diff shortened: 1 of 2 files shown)


poteto:
@homu r+

addon/helpers/pipe.js
```diff
@@ -12,7 +12,7 @@ export function pipe(actions = []) {
         return curr(...args);
```
cowboyd: I also just noticed this: If any function in the pipe returns `undefined`, the next function will receive the original arguments. I don't know if this is a problem, but it could lead to subtle bugs around the ordering of operations. Maybe use a unique tag instead of `undefined`?

```js
const first = new Object();

export function pipe(actions = []) {
  return function(...args) {
    return actions.reduce((acc, curr) => {
      if (acc === first) {
        return curr(...args);
      } 
      
      return curr(acc);
    }, first);
  };
}
```

poteto: Good point, I could also use the `currentIndex` from the `reduce` instead

PR #81, opened 2016-02-24
Remove unused methods and semicolon

poteto:


martndemus:
I wonder why JSHint didn't catch that?

poteto:
Good question

poteto:
@homu r+

Issue #54, opened 2016-02-22
Discuss argument ordering


martndemus:
```hbs
{{! target object first }}
{{take (sort (filter-by array "active") "lastName" "firstName") 5}}

{{! target object last }}
{{take 5 (sort "lastName" "firstName" (filter-by "active" array))}}
```

Target object last looks like a clear winner to me.

poteto:
One potential point for confusion with target object last might be the discrepancy between the Ember method signature and the helper signature:

```js
computed.mapBy('array', 'prop')
```

vs

```hbs
{{map-by "prop" array}}
```

martndemus:
I think making exceptions to reflect existing Ember API should be considered, but we should not end up with an inconsistent API because of making those exceptions. We should really stick with one option.

bcardarella:
A convention in Elixir is that the "subject" is always the first argument. This creates a contract between functions so that they can always be used with composability in the same way. If you were to take this lesson to heart it would prefer the 2nd version

poteto:
Confirm, the goal for this issue is to decide on the contract moving forward, but to also be aware of the potential downsides. I think if we document the order of arguments clearly in the README we could avoid the confusion between the Ember method and the helper.

bcardarella:
Considering that some functions might only take one argument I think the first argument style would be the easiest to conform to.

WRT readability, the example above `take 5 ...` reads more clearly but I don't think this should be the example to develop this standard around

martndemus:
Good thing to consider: actions are also in the scope of this addon. Actions will greatly benefit from having the subject as last param, because of the currying.

brzpegasus:
I prefer this format:

```hbs
{{take (sort (filter-by array "active") "lastName" "firstName") 5}}
```

Particularly with sorting and filtering, I like to see the collection that is going to be acted upon first, followed by the property names to sort or filter by.

flexyford:
The second option is better for currying, as @martndemus mentioned. 

Looking at `take` from a FP perspective, let's define `take5`, which will return the first 5 items from a list. This makes it more composable in our templates.

```javascript
// js example for clarity
let nums = [1,2,3,4,5,6,7,8,9]
var take5 = take(5) //=> function(list)
take5(nums) //=> [1,2,3,4,5]
```

```hbs
{{!hbs example}}
{{#with (take 5) as |take5|}}
  Active:
  {{take5 (sort (filter-by array "active") "lastName" "firstName")}}
  Inactive:
[...]

PR #86, opened 2016-02-25
Fix argument order in `map-by` example

HeroicEric:
Update example to match changes resulting from https://github.com/DockYard/ember-composable-helpers/issues/54


poteto:
@homu r+

poteto:
Thanks @HeroicEric! :)

Issue #32, opened 2016-02-19
Flatten helper


poteto:
Deep flatten. 

```js
flatten([1], [2], [3]) === [1, 2, 3];
flatten([1], [2, 3, [4, 5]]) === [1, 2, 3, 4, 5];
```

```hbs
{{#each (flatten (repeat 3 (range 1 3 true))) as |number|}}
  {{! renders something like `1 2 3 1 2 3 1 2 3`}}
{{/each}}
```

```hbs
{{#each (flatten (repeat 3 (take 3 items))) as |number|}}
  {{! renders something like `first second third first second third first second third`}}
{{/each}}
```

zigahertz:
it looks @vikram7 had a PR for this, but it got closed; is there any plan to keep working on that branch? I could use a `flatten` helper myself...

martndemus:
@zigahertz feel free to open a fresh PR!

Asherlc:
See https://github.com/DockYard/ember-composable-helpers/pull/158

martndemus:
Implemented by #158

PR #96, opened 2016-03-01
Adds flatten helper

vikram7:
This PR adds the `flatten`helper per https://github.com/DockYard/ember-composable-helpers/issues/32

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ For action helpers, this will mean better currying semantics:
   + [`append`](#append)
   + [`chunk`](#chunk)
   + [`without`](#without)
+  + [`flatten`](#flatten)
 * [Object](#object-helpers)
   + [`group-by`](#group-by)
 * [Math](#math-helpers)
@@ -478,6 +479,17 @@ Returns the given array without the given item(s).
 ```hbs
 {{#each (without selectedItem items) as |remainingItem|}}
   {{remainingItem.name}}
+{{/each}}
+```
+
+**[⬆️ back to top](#available-helpers)**
+
+#### `flatten`
+Flattens an array to a single dimension.
+
+```hbs
+{{#each (flatten anArrayOfNamesWithMultipleDimensions) as |name|}}
+  Name: {{name}}
 {{/each}}
 ```
 
--- a/addon/helpers/flatten.js
+++ b/addon/helpers/flatten.js
@@ -1,14 +1,15 @@
 import Ember from 'ember';
 
 const {
-  Helper: { helper }
+  Helper: { helper },
+  typeOf
 } = Ember;
 
 export function flatten(array) {
   array = array || [];
 
   return array.reduce((flat, toFlatten) => {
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -71,21 +71,22 @@ For action helpers, this will mean better currying semantics:
   + [`underscore`](#underscore)
   + [`w`](#w)
 * [Array](#array-helpers)
-  + [`map-by`](#map-by)
```
poteto: Please don't sort, you can do it as a separate PR.

addon/helpers/flatten.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+const {
+  Helper: { helper }
+} = Ember;
+
+export function flatten(array) {
+  array = array || [];
+
+  return array.reduce((flat, toFlatten) => {
```
poteto: This helper will not recompute when an array has changes. Please see something like the `contains` helper for an example.

tests/integration/helpers/flatten-test.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('it handles null input', function(assert) {
+  this.render(hbs`
+    {{#each (flatten null) as |number|}}{{number}}{{/each}}
+  `);
+
+  assert.equal(this.$().text().trim(), '', 'should handle null input');
+});
+
+test('it handles undefined input', function(assert) {
```
poteto: Needs another test for array changes

addon/helpers/flatten.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+const {
+  Helper: { helper }
+} = Ember;
+
+export function flatten(array) {
+  array = array || [];
```
poteto: Can we have this as `array || [array];`? I would prefer that this method always returns an array regardless of the value passed into it.

Also, we should be using an `isArray` check here. It is still valid if the array contains `null` or `undefined` values.

poteto:
If you rebase this against master, `beta` will be an allowed failure.

tests/integration/helpers/flatten-test.js
vikram7: :+1: Has been added above (line 25).

addon/helpers/flatten.js
```diff
@@ -0,0 +1,40 @@
...
+  isArray,
+  observer,
+  set
+} = Ember;
+
+export function flatten(array) {
+  array = array || [array];
+
+  return array.reduce((flat, toFlatten) => {
+    return flat.concat(Array.isArray(toFlatten) ? flatten(toFlatten) : toFlatten);
```
poteto: Can we use `Ember.typeOf` here? Just so that other kinds of Ember Arrays will also be true here.

taras:
@vikram7 @poteto what happened to this PR? Should we re-born it?

martndemus:
@taras this got stale and apparently the OP closed it. Feel free to open a fresh PR!

PR #99, opened 2016-03-05
Make `pipe` helper promise aware

### LINKED ISSUE #85: Promise-aware pipe helper
- [x] Check if the return value is a Promise
- [x] Promise should yield a value in its `then` hook, which is then passed into the next func 
- [x] If the Promise rejects, we abort the entire chain

poteto:
Closes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,8 @@ Pipes the return values of actions in a sequence of actions. This is useful to c
 </button>
 ```
 
+The `pipe` helper is also Promise-aware, meaning that if any action in the pipeline returns a Promise, its return value will be piped into the next action.
+
 **[⬆️️ back to top](#available-helpers)**
 
 #### `compute`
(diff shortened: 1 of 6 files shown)


poteto:
@homu r+

henrymazza:
If promise is rejected shouldn't it halt the chain? I'm thinking on scenarios like
```
<button {{action (pipe addToCart purchase redirectToThankYouPage) item}}>
  1-Click Buy
</button>
```
if `addToCart` returns an error (e.g. connection lost) then purchase shouldn't be hit.

henrymazza:
`isPromise()` should test for `PromiseProxy` too.

poteto:
Yes, it already does that. I added a test to make it more obvious – #101

henrymazza:
Found the problem here. My function returns a PromiseProxy which gets false on `isPromise()`. Perhaps test for that too?

poteto:
@henrymazza Good catch! Can you please open a new issue for it?

henrymazza:
Ok. I wish I knew how to fix it, but I couldn't find how do detect the Mixin. `instanceof` doesn't work.

henrymazza:
#102

PR #103, opened 2016-03-05
Make `pipe` work with PromiseProxies

### LINKED ISSUE #102: isPromise should test for other categories of thennable
Currently utils module `isPromise` tests only for `instanceof` Promise, however in Ember the occurrence of PromiseProxy Mixin is far more comum. I can't recall right now, but may be other types of thennable objects besides those two, so perhaps isPromise should be chnged to isThennable? Ideas?

poteto:
- Relaxed the `instanceof` check to duck type if the object implements
  Promise-like methods to work with Objects that include the
  PromiseProxy mixin

Closes #102

### DIFF
--- a/addon/utils/is-promise.js
+++ b/addon/utils/is-promise.js
@@ -1,7 +1,13 @@
 import Ember from 'ember';
 
-const { RSVP: { Promise }, isPresent } = Ember;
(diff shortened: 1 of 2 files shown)


poteto:
@homu r+

PR #108, opened 2016-03-07
adds uniq helper

zigahertz:
I thought this would be helpful to have, I needed it in one of the projects I'm working on.

### DIFF
--- a/addon/helpers/uniq.js
+++ b/addon/helpers/uniq.js
@@ -1,0 +1,29 @@
+import Ember from 'ember';
+
+const {
+  Helper,
+  get,
+  set,
+  observer,
+  isArray,
+  A: emberArray,
+  computed: { uniq }
+} = Ember;
+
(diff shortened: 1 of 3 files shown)


martndemus:
@zigahertz Uniq is an alias for `union`, we already have that helper. 

Maybe add an alias here too? Thoughts @poteto?

poteto:
I think I would prefer to have one helper, but to update the README so that people looking for `uniq` will be pointed to use `union` instead.

zigahertz:
@martndemus @poteto thanks! I didn't realize that `union` had this functionality for a single array. I can submit a PR that updates the README for this purpose.

martndemus:
Yes please!

zigahertz:
alright, I've done so at https://github.com/DockYard/ember-composable-helpers/pull/113. Closing this PR.

PR #116, opened 2016-03-11
Add `Without` helper

### LINKED ISSUE #14: Without helper
Without a single item:

```js
without([foo, bar, baz], baz) === [foo, bar];
```

```hbs
{{#each (without selectedItem items) as |remainingItem|}}
  {{remainingItem.name}}
{{/each}}
```

Without a collection of items (array):

```js
without([foo, bar, baz], [foo, bar]) === [baz];
```

```hbs
{{#each (without selectedItems items) as |remainingItem|}}
  {{remainingItem.name}}
{{/each}}
```

poteto:
Closes #14

### DIFF
--- a/tests/integration/helpers/without-test.js
+++ b/tests/integration/helpers/without-test.js
@@ -1,0 +1,43 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('without', 'Integration | Helper | {{without}}', {
+  integration: true
+});
+
+test('it returns a new array with given value ommitted', function(assert) {
+  this.set('items', ['foo', 'bar', 'baz']);
+
+  this.render(hbs`
+    {{~#each (without "foo" items) as |remaining|~}}
+      {{remaining}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'barbaz', 'should render remaining values');
+});
+
+test('it returns a new array with given values ommitted', function(assert) {
+  this.set('items', ['foo', 'bar', 'baz']);
+  this.set('selectedItems', ['foo', 'bar']);
+
+  this.render(hbs`
+    {{~#each (without selectedItems items) as |remaining|~}}
+      {{remaining}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'baz', 'should render remaining values');
+});
+
+test('it returns the same array when no values are ommitted', function(assert) {
+  this.set('items', ['foo', 'bar', 'baz']);
+
+  this.render(hbs`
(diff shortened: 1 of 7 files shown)


martndemus:
:shipit:

poteto:
I hope to be able to refactor some of the other Helpers with private "Helper factories" in a future PR

tests/integration/helpers/without-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+moduleForComponent('without', 'Integration | Helper | {{without}}', {
+  integration: true
+});
+
+test('it returns a new array with given value ommitted', function(assert) {
+  this.set('items', ['foo', 'bar', 'baz']);
+
+  this.render(hbs`
+    {{~#each (without 'foo' items) as |remaining|~}}
```
courajs: What are these ~? I've never seen them in a template before

tests/unit/helpers/without-test.js
```diff
@@ -0,0 +1,18 @@
...
+
+  assert.deepEqual(result, expectedResult, 'should return the remaining values');
+});
+
+test('it returns the array with the given values ommitted', function(assert) {
+  let expectedResult = [78];
+  let result = without([42, 56], [42, 56, 78]);
+
+  assert.deepEqual(result, expectedResult, 'should return the remaining values');
+});
```
courajs: Maybe add a test for when the given value is not in the given array?

tests/integration/helpers/without-test.js
poteto: It strips whitespace!

tests/unit/helpers/without-test.js
poteto: Good call

addon/-private/create-reducer-helper.js
```diff
@@ -0,0 +1,41 @@
...
+  computed,
+  observer,
+  set,
+  get
+} = Ember;
+
+/**
+ * Creates a generic Helper class implementation that expects a `needle` and
+ * `haystack` as arguments. A `reducer` function is required to be passed in
+ * that reduces the `haystack` down into a single value.
```
courajs: I would say this file doesn't actually have anything to do with reduce. It actually just creates a helper that takes two array arguments, and updates when their contents change. The two times you use it, you happen to use a reducing function. It can be left as is, but it's a little confusing because the names introduce a concept that doesn't exist in the code. I was confused and looking around for a `reduce` call

poteto: It's a good point, I was struggling to come up with a name for it because of how generic it is...

courajs: yeah... `array-observing-helper`?

courajs:
LGTM

poteto:
@homu r+

PR #134, opened 2016-04-12
Add queue helper

skeate:
Similar to the pipe helper, but instead of chaining return values of one action as the arguments of another, this will pass the same arguments to all actions.

Not sure if this is something you're interested in, but I found it useful for a few situations, like chaining a helper function in with the real business function, but without having to make the helper function pass its args along.

Also not sure about the name. Some other possible names: `fork`, `multi`, `do-each`. Open to other suggestions as well.

### DIFF
--- a/addon/helpers/queue.js
+++ b/addon/helpers/queue.js
@@ -1,0 +1,24 @@
+import { helper } from 'ember-helper';
+import isPromise from '../utils/is-promise';
+
+export function queue(actions = []) {
+  return function(...args) {
+    const invokeWithArgs = function(acc, curr) {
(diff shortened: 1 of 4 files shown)


skeate:
I am guessing the test failure is related to https://github.com/emberjs/ember.js/issues/13050

taras:
@skeate this PR has been sitting here since Apr 12th. I'm not sure if it's a great fit here, but I'd welcome this helper in `ember-reactive-helpers`. 

What is the result of the execution of this helper?

martndemus:
Merged in #214

martndemus:
Released in `1.1.0`

PR #138, opened 2016-04-19
Updates `filter-by` documentation in README.

patrickberkeley:
Updates `filter-by` documentation in README to reference the correct argument that's passed in as a value.


poteto:
@homu r+

PR #140, opened 2016-04-22
Add next and previous helpers

### LINKED ISSUE #62: Next, Previous helper
```js
next([1, 2, 3, 4, 5], 2) === 3;
next([1, 2, 3, 4, 5], 3) === 4;
previous([1, 2, 3, 4, 5], 3) === 2;
previous([1, 2, 3, 4, 5], 4) === 3;
```

```hbs
<button onclick={{action (mut selectedItem) (previous items selectedItem)}}>Previous</button>
<button onclick={{action (mut selectedItem) (next items selectedItem)}}>Next</button>
```

Optionally also add `hasNext` and `hasPrevious`.

maabernethy:
* Includes `hasNext` and `hasPrevious`
* Adds `addIndex` and `isEqual` util

Closes #62

### DIFF
--- a/addon/-private/create-needle-haystack-helper.js
+++ b/addon/-private/create-needle-haystack-helper.js
@@ -5,7 +5,7 @@ import get from 'ember-metal/get';
 import observer from 'ember-metal/observer';
 import set from 'ember-metal/set';
 
-const { K } = Ember;
+const { K, isEmpty } = Ember;
 
 /**
  * Creates a generic Helper class implementation that expects a `needle` and
@@ -18,16 +18,23 @@ const { K } = Ember;
  */
 export default function createNeedleHaystackHelper(fn = K) {
   return Helper.extend({
-    content: computed('needle.[]', 'haystack.[]', function() {
+    content: computed('needle.[]', 'haystack.[]', 'option', function() {
       let needle = get(this, 'needle');
       let haystack = get(this, 'haystack');
+      let option = get(this, 'option');
 
-      return fn(needle, haystack);
+      return fn(needle, haystack, option);
     }).readOnly(),
 
-    compute([needle, haystack]) {
+    compute([needle, option, haystack]) {
+      if (isEmpty(haystack)) {
+        haystack = option;
+        option = null;
+      }
+
       set(this, 'needle', needle);
       set(this, 'haystack', haystack);
+      set(this, 'option', option);
 
       return get(this, 'content');
     },
--- a/addon/helpers/has-previous.js
+++ b/addon/helpers/has-previous.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+import { previous } from './previous';
+import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
+import isEqual from '../utils/is-equal';
+
+const { isPresent } = Ember;
+
+export function hasPrevious(currentValue, array, useDeepEqual = false) {
+  let previousValue = previous(currentValue, array, useDeepEqual);
+  let isNotSameValue = !isEqual(previousValue, currentValue, useDeepEqual);
+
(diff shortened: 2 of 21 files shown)


addon/helpers/has-previous.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+import { previous } from './previous';
+import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
+
+let { isPresent } = Ember;
```
poteto: `const`? 👼

addon/helpers/next.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+import getIndex from '../utils/get-index';
+import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
+
+let { isPresent } = Ember;
+
+export function next(array, currentValue) {
+  let currentIndex = getIndex(array, currentValue);
+  return isPresent(currentIndex) ? array[currentIndex + 1] : null;
```
poteto: In the event that the `currentValue` is the last item in the array, I think we would want to return the item instead of `undefined`

addon/helpers/previous.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+import getIndex from '../utils/get-index';
+import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
+
+let { isPresent } = Ember;
+
+export function previous(array, currentValue) {
+  let currentIndex = getIndex(array, currentValue);
+  return isPresent(currentIndex) ? array[currentIndex - 1] : null;
```
poteto: Same here, I think we would want to return the same item if it's already the first item in the array

addon/utils/get-index.js
```diff
@@ -0,0 +1,14 @@
+import isObject from '../utils/is-object';
+
+export default function getIndex(array, currentValue) {
+  let index;
+  if (isObject(currentValue)) {
+    index = array.findIndex((object) => {
+      return JSON.stringify(object) === JSON.stringify(currentValue);
```
poteto: could we pass in an option to this function, say `deepEqual`? It can default to `false`, then we only need to compare the reference using `===` and stringify if set to `true`

poteto:
Looks pretty good, just minor comments! I would also rename `has{Next,Previous}` to `has-{next,previous}` 

Thanks for this :)

addon/helpers/has-previous.js
maabernethy: whoops 😬

addon/helpers/next.js
maabernethy: 👍 Yeah I wasn't sure what the ideal/ most useful thing to return was

addon/utils/get-index.js
maabernethy: 👍

maabernethy: @poteto Would you prefer `useDeepEqual` be a named param? Or just a normal one

maabernethy:
This has been updated per comments. Let me know if you would prefer `useDeepEqual` to be a named param

addon/-private/create-needle-haystack-helper.js
```diff
@@ -18,16 +18,18 @@ const { K } = Ember;
...
       let needle = get(this, 'needle');
       let haystack = get(this, 'haystack');
+      let option = get(this, 'option');
 
-      return fn(needle, haystack);
+      return fn(needle, haystack, option);
     }).readOnly(),
 
-    compute([needle, haystack]) {
+    compute([needle, haystack, option]) {
```
martndemus: I think `option` needs to go before `haystack`. The subject where the helper is operating on should always be last.

martndemus:
LGTM except that the argument order is inconsistent with the other helpers.

PR #141, opened 2016-04-25
[BUGFIX] {{filter-by}} must also observe changes in the value

cibernox:
At the moment `{{filter-by 'name' userName users}}` do not update if `userName` is updated.


martndemus:
@homu r+

cibernox:
👍

martndemus:
Looks like @poteto :ship: this with version `0.22.1`!

PR #143, opened 2016-04-26
adds array helper and tests

### LINKED ISSUE #142: (array ) helper maybe?
Sometimes it's helpful to compose an array in the template. 

What do you think about something like this?

```hbs
{{#each (array 'foo' 'bar' 'baz') as |word|}}
  {{word}}
{{/each}}
```

The implementation is like hash helper,

```javascript
export default Ember.Helper.helper(function(params) {
  return params;
});
```

If you agree, I'll make a PR. 

Also, if there is another way to do with that's already built-in, please let me know!

zigahertz:
closes #142 

Signed-off-by: Taras Mankovski <taras@embersherpa.com>

### DIFF
--- a/addon/helpers/array.js
+++ b/addon/helpers/array.js
@@ -1,0 +1,8 @@
+import Ember from 'ember';
+import { A as emberArray } from 'ember-array/utils';
+
+export function array(params) {
+  return emberArray(params);
+}
+
+export default Ember.Helper.helper(array);
--- a/tests/integration/helpers/array-test.js
+++ b/tests/integration/helpers/array-test.js
@@ -1,0 +1,46 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('array', 'Integration | Helper | {{array}}', {
+  integration: true
+});
+
+test('creates array object in template', function(assert) {
+  this.render(hbs`
+    {{~#each (array 1 2 3) as |number|~}}
+      {{number}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '123', 'creates array object in template');
+});
+
+test('has length property', function(assert) {
+  this.render(hbs`
+    {{~#with (array 1 2 3 5) as |a|~}}
+      {{a.length}}
+    {{~/with~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '4', 'length is accessible');
+});
+
+test('re-evaluates when parameter changes', function(assert) {
+  this.render(hbs`
+    {{~#each (array 1 2 dynamic 3) as |number|~}}
+      {{number}}
+    {{~/each~}}
+  `);
+
+  assert.equal(this.$().text().trim(), '123', 'dynamic is not included when undefined');
+
+  this.set('dynamic', 'π');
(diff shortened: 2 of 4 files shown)


addon/helpers/array.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export function array(params=[]) {
+  return params;
```
poteto: If you wrap this with `Ember.A()`, does it still appear truthy in the template when empty?

addon/helpers/array.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+const { A, isNone };
+
+export function array(params) {
+  if (isNone(params)) {
+    params = new A();
```
zigahertz: @poteto it doesn't make a difference - but we added it here

addon/helpers/array.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+const { A, isNone };
```
poteto: You can import it instead:

```js
import { A as emberArray } from 'ember-array/utils';
```

addon/helpers/array.js
```diff
@@ -0,0 +1,12 @@
...
+
+export function array(params) {
+  if (isNone(params)) {
```
poteto: This check isn't necessary, you can just return:

```js
return emberArray(params);
```

tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,58 @@
...
+  this.set('dynamic', 'π');
+  assert.equal(this.$().text().trim(), '12π3', 'dynamic is included');
+});
+
+test('if evaluates empty array to false', function(assert) {
+  this.render(hbs`
+    {{~if (array) 'true' 'false'~}}
+  `);
+  assert.ok([], 'JavaScript evaluates empty array as true');
+  assert.equal(this.$().text().trim(), 'false', 'empty array evaluates to false');
```
poteto: I think this behavior is fine – it makes sense when it is in the context of a template. An empty array _should_ be falsey when used in hbs

tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  this.set('dynamic', 'π');
+  assert.equal(this.$().text().trim(), '12π3', 'dynamic is included');
+});
+
+test('if evaluates empty array to false', function(assert) {
+  this.render(hbs`
+    {{~if (array) 'true' 'false'~}}
+  `);
+  assert.ok([], 'JavaScript evaluates empty array as true');
```
poteto: This is not a useful assertion, we can remove it

tests/integration/helpers/array-test.js
```diff
@@ -0,0 +1,58 @@
...
+  assert.equal(this.$().text().trim(), 'false', 'empty array evaluates to false');
+});
+
+test('if helper treats empty arrays as false', function(assert) {
```
poteto: I think this test is outside the scope of this helper and can be removed

zigahertz:
@poteto thanks for your comments, just updated the code to reflect them.

poteto:
Looks great @zigahertz! Could you please squash it down into a single commit? Thank you!

zigahertz:
thanks @poteto, squash is done!

poteto:
@homu r+

PR #148, opened 2016-04-28
Allow chunk helper to work with CP length.

### LINKED ISSUE #147: chunk helper does not respect computed length properties
When passing the result of `Store.peekAll` (a `RecordArray`), length is a computed property. The chunk helper does not `get` the computed property and throws an exception.

[Here](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/chunk.js#L13) we don't `.get` the length from the array, which causes the length var to be assigned to a CP. Then, [this line](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/chunk.js#L20) throws an exception since length isn't an int:

```
chunk.js:45 Uncaught RangeError: Invalid array length
```

blimmer:
Fixes #147.

### DIFF
--- a/addon/helpers/chunk.js
+++ b/addon/helpers/chunk.js
@@ -10,7 +10,11 @@ const { max, ceil } = Math;
 export function chunk(num, array) {
   let integer = parseInt(num, 10);
   let size = max(integer, 0);
-  let length = isArray(array) ? array.length : 0;
+
+  let length = 0;
+  if (isArray(array)) {
+    length = get(array, 'length');
+  }
 
   if (!length || size < 1) {
     return [];
--- a/tests/unit/helpers/chunk-test.js
+++ b/tests/unit/helpers/chunk-test.js
@@ -1,3 +1,6 @@
+import Ember from 'ember';
+import { A as emberArray } from 'ember-array/utils';
+import computed from 'ember-computed';
 import { chunk } from 'dummy/helpers/chunk';
 import { module, test } from 'qunit';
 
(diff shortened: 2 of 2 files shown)


addon/helpers/chunk.js
```diff
@@ -10,7 +10,11 @@ const { max, ceil } = Math;
 export function chunk(num, array) {
   let integer = parseInt(num, 10);
   let size = max(integer, 0);
-  let length = isArray(array) ? array.length : 0;
+
+  let length = 0;
+  if (isArray(array)) {
+    length = typeof array.get === 'function' ? array.get('length') : array.length;
```
martndemus: `get(array, 'length')` will work for both cases!

tests/unit/helpers/chunk-test.js
```diff
@@ -54,3 +55,16 @@ test('it chunks an array into parts of less than current array length', function
   assert.deepEqual(result1, expectedResult1, 'should return array of chunked arrays');
   assert.deepEqual(result2, expectedResult2, 'should return array of chunked arrays');
 });
+
+test('it handles arrays with computed property lengths', function(assert) {
+  let ap = Ember.ArrayProxy.extend({
+    content: Ember.A([1, 2, 3]),
+    length:  Ember.computed('content.[]', function() {
```
martndemus: Can you import `Ember.A` and `Ember.computed` as seen in other helpers?

martndemus: See: 
- https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/filter-by.js#L2
- https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/chunk.js#L2

martndemus:
Awesome!
@homu r+

martndemus:
@blimmer published in 0.23.1

PR #151, opened 2016-04-28
Fix sort-by with a function

Asherlc:
Ember.computed.sort seems to require you to send in a raw function (not a key that references a function) for it to work.


martndemus:
Can you add a test that fails without the fix?

Asherlc:
This may be a silly question, but how do I run tests for an addon?

martndemus:
`ember test` :)

Asherlc:
Ok, at 65a9649830efe818f90daa93c03a580cdef3741f there is a test which fails with a function, but passes in 084e7503cc555289eda674bbf08b44709aaa727b

Does that do the trick?

Asherlc:
checking in on this

martndemus:
@Asherlc can you add a little documentation to README.md?

Asherlc:
sure thing, done

martndemus:
Closed via #191

PR #159, opened 2016-05-15
length helper

Asherlc:
Returns the length of an array


martndemus:
1. Why is this better than `{{array.length}}` ?
2. `app/helpers/length.js` should only re-export from `addon/helpers/length.js`
3. Needs test that tests for changes to array
4. Needs entry in `addon/index.js`

Asherlc:
I was going to say this could be composed with other helpers vs `array.length`, but it occurs to me that you could just use the `get` helper, so it's redundant.

PR #160, opened 2016-05-15
slice helper

Asherlc:
Slices an array


martndemus:
1) I'm not particularly sold on the hash param addition
2) Needs a test for when the array contents are updated, as I suspect this will not update when the content of the array updates.

Asherlc:
1. I've removed the hash params
2. You were correct -- this has been remedied.
3. Done

Asherlc:
Merge conflicts resolved

poteto:
Looks like there's some overlap with the [`take`](https://github.com/DockYard/ember-composable-helpers/blob/master/app/helpers/take.js) helper, any chance you could DRY it up?

martndemus:
Closed via #190

PR #172, opened 2016-06-04
Implementing object-at helper.

gmurphey:
The goal of this helper was to facilitate pagination in conjunction with the `chunk` helper. You can get similar results with Ember's `get` helper, but it requires the index to be a string, as far as I can tell.

### DIFF
--- a/addon/helpers/object-at.js
+++ b/addon/helpers/object-at.js
@@ -1,0 +1,36 @@
+import Helper from 'ember-helper';
+import { A as emberArray, isEmberArray as isArray } from 'ember-array/utils';
+import computed from 'ember-computed';
+import observer from 'ember-metal/observer';
+import get from 'ember-metal/get';
+import set from 'ember-metal/set';
+
+export function objectAt(index, array) {
+  if (!isArray(array)) {
+    return undefined;
+  }
+
+  index = parseInt(index, 10);
+
+  return emberArray(array).objectAt(index);
+}
+
+export default Helper.extend({
+  content: computed('index', 'array.[]', function() {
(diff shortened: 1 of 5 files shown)


martndemus:
Hi @gmurphey,

Overall this looks like a nice addition!

There are a few things this PR needs before it can be merged:
- [ ] Needs an entry in `addon/index.js`
- [ ] Needs a test to see what happens when an element is inserted into the array at the given position

addon/helpers/object-at.js
```diff
@@ -0,0 +1,10 @@
+import { helper } from 'ember-helper';
+import { A as emberArray } from 'ember-array/utils';
+
+export function objectAt(params) {
```
martndemus: `export function objectAt([index, array]) {`

addon/helpers/object-at.js
```diff
@@ -0,0 +1,10 @@
...
+  let index = parseInt(params[0], 10);
```
martndemus: `let index = parseInt(index, 10);`

addon/helpers/object-at.js
```diff
@@ -0,0 +1,10 @@
...
+
+  return emberArray(params[1]).objectAt(index);
```
martndemus: `return emberArray(array).objectAt(index);`

gmurphey:
Thanks for the feedback, @martndemus! Should be all set.

martndemus:
Fantastic!
@poteto r?

poteto:
Lgtm!

martndemus:
@homu r+

PR #192, opened 2016-07-22
[BUGFIX] {{filter-by}} must work without prototype extensions

cibernox:


tests/integration/helpers/find-by-test.js
```diff
@@ -24,6 +24,22 @@ test('It finds a value by a property', function(assert) {
...
 
+test('It finds a value by a property in arrays without prototype extensions', function(assert) {
+  this.set('array', [
+    { foo: true, name: 'a' },
+    { foo: false, name: 'b' },
+    { foo: true, name: 'c' }
+  ]);
+
+  this.render(hbs`
+    {{~#with (find-by 'name' 'b' array) as |item|~}}
```
martndemus: Use double-quotes in HBS

martndemus: Ugh, the whole file is this way :(
I'll fix in separate PR.

addon/helpers/find-by.js
```diff
@@ -29,7 +29,7 @@ export default Helper.extend({
       let array = get(this, 'array');
       let value = get(this, 'value');
 
-      return array.findBy(byPath, value);
+      return array.find(e => get(e, byPath) === value);
```
martndemus: I think it's preferable to wrap `array` in `emberArray` and then just use `findBy`.

cibernox: Out of curiosity, why is that preferable? (Unless this addons targets 1.13 and ie8)

martndemus: Because IMO a `findBy` reads better then a terse arrow function.

martndemus:
Fixed by #195

Issue #197, opened 2016-07-28
Inconsistent checks for arrays


jondayton:
In checking to see if the object is an array

`contains` uses `typeOf(haystack) !== 'array')`
`invokes` uses `isEmberArray(obj)`

This can lead to inconsistencies in using Ember arrays in templates.

martndemus:
Would you be willing to shoot a PR that updates all cases to `isEmberArray`?

jondayton:
Yes I'll do that.

martndemus:
You're awesome! :+1:

jondayton:
I'm noticing there's also `{ isEmberArray as isArray }`

is the preferred syntax `isEmberArray(array)` or `isArray(array)`?

martndemus:
It should all be `isEmberArray`.

PR #198, opened 2016-07-28
standardize array checks to Ember arrays (#197)

jondayton:
There were three different syntaxes being used:
```js
import { typeOf } from 'ember-utils';
typeOf(array) !== 'array';
```
```js
import { isEmberArray as isArray } from 'ember-array/utils';
isArray(array);
```
```js
import { isEmberArray } from 'ember-array/utils';
isEmberArray(array);
```

This standardizes to use the third syntax across all helpers.

### DIFF
--- a/addon/-private/create-multi-array-helper.js
+++ b/addon/-private/create-multi-array-helper.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import {
   A as emberArray,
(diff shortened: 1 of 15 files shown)


poteto:
LGTM :shipit:

poteto:
@homu r+

PR #203, opened 2016-08-10
[Docs] useDeepEqual is last parameter

deepflame:
Hi!

If I understand the source correctly `useDeepEqual` is the last parameter. In the README it was the second...

e.g. Source:

```javascript
export function next(currentValue, array, useDeepEqual = false) {
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,7 +687,7 @@ Returns the next element in the array given the current element. **Note:** Accep
 parameter, `useDeepEqual`, to flag whether a deep equal comparison should be performed.
 
 ```hbs
-<button onclick={{action (mut selectedItem) (next selectedItem useDeepEqual items)}}>Next</button>
+<button onclick={{action (mut selectedItem) (next selectedItem items useDeepEqual)}}>Next</button>
 ```
 
 **[⬆️ back to top](#available-helpers)**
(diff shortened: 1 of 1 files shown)


martndemus:
Then this is a bug in the code, `items` should be the last param. Can you adjust your PR to fix the code instead of the docs?

deepflame:
ok found it. Was pretty confusing.

It is the `create-needle-haystack-helper`. It switches the order of the parameters in [line 26](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/-private/create-needle-haystack-helper.js#L26) and [line 29](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/-private/create-needle-haystack-helper.js#L29).

Any reasons the options are in the middle? At least I would prefer having the options at the end like in the method implementations.

martndemus:
> Any reasons the options are in the middle?
Because: https://github.com/DockYard/ember-composable-helpers#argument-ordering

deepflame:
yeah, but I see that you have the list at the end for the other helpers as well. So I guess this is just about consistency.

So, would you rather like to have the implementations have the same method signature as they are called to lessen confusion or should I just leave it as is?

martndemus:
@deepflame Leave as it is, unless it doesn't work as specified in the README right now.

deepflame:
No, everything good. Thanks for the clarification

PR #205, opened 2016-08-10
`cycle` array helper

deepflame:
this is like the `next` array helper but starts again at the beginning once the last element is reached.


martndemus:
Instead of adding a new helper, could you modify next and previous to accept a flag that makes it repeat?

PR #207, opened 2016-08-15
Upgrade Ember CLI to 2.7.0 + Suave'd

nicomihalich:
Notes:

- Bumped `ember-exam` to `v0.4.6` for [this change](https://github.com/trentmwillis/ember-exam/commit/667363ff792acb2770d3c8b48bb14188d8cc9526)

- Bumped `ember-suave` to `v4.0.0` for [this change](https://github.com/DockYard/ember-suave/commit/7ac090da21141837776aefb99fd8504e5c9c5648)

### DIFF
--- a/addon/helpers/contains.js
+++ b/addon/helpers/contains.js
@@ -2,9 +2,10 @@ import { A as emberArray } from 'ember-array/utils';
 import get from 'ember-metal/get';
 import { isEmberArray } from 'ember-array/utils';
 import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
+import includes from '../utils/includes';
 
 function _contains(needle, haystack) {
-  return emberArray(haystack).contains(needle);
+  return includes(emberArray(haystack), needle);
 }
 
 export function contains(needle, haystack) {
--- a/package.json
+++ b/package.json
@@ -27,46 +27,48 @@
     "broccoli-asset-rev": "^2.4.2",
     "chai": "^3.5.0",
     "ember-ajax": "^2.0.1",
-    "ember-cli": "2.6.2",
+    "ember-cli": "2.7.0",
     "ember-cli-app-version": "^1.0.0",
     "ember-cli-dependency-checker": "^1.2.0",
     "ember-cli-htmlbars": "^1.0.3",
     "ember-cli-htmlbars-inline-precompile": "^0.3.1",
     "ember-cli-inject-live-reload": "^1.4.0",
     "ember-cli-jshint": "^1.0.0",
-    "ember-cli-qunit": "^1.4.0",
+    "ember-cli-qunit": "^2.0.0",
     "ember-cli-release": "^0.2.9",
     "ember-cli-sri": "^2.1.0",
+    "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
+    "ember-data": "^2.7.0",
     "ember-disable-prototype-extensions": "^1.1.0",
     "ember-disable-proxy-controllers": "^1.0.1",
     "ember-each-in-polyfill": "0.1.3",
-    "ember-exam": "0.4.4",
+    "ember-exam": "0.4.6",
     "ember-export-application-global": "^1.0.5",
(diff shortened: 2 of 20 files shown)


poteto:
Circle's chrome seems pretty outdated

addon/helpers/contains.js
```diff
@@ -4,7 +4,7 @@ import { isEmberArray } from 'ember-array/utils';
 import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
 
 function _contains(needle, haystack) {
-  return emberArray(haystack).contains(needle);
+  return emberArray(haystack).includes(needle);
```
poteto: Might want to still use `contains`, isn't `includes` new? This will break old versions of Ember.

martndemus:
@poteto Yes, but using includes is maybe not the smartest idea, has no IE support. We should try to comply with IE9+, just like Ember itself.

nicomihalich:
Looks like Circle is complaining about `findIndex` as well.  I didn't touch that here though.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

nicomihalich:
@martndemus @poteto What do you think of the polyfill solution https://github.com/DockYard/ember-composable-helpers/pull/207/commits/889192a8a8081c9ff642cfb9880a2f5162650e84 ?  I can do similar for `findIndex` to make the build pass on Circle

poteto:
`findIndex` was added in Chrome 45.0, Circle is using 43.0. Can we bring back the chrome upgrade? We shouldn't need to polyfill this.

nicomihalich:
Neither of them work in IE though, do we need to support that?  If not I can just get rid of both polyfills and try to get the chrome update working.  It wasn't playing nice before: https://circleci.com/gh/DockYard/ember-composable-helpers/13

poteto:
Try this:

```yml
machine:
  node:
    version: "stable"
dependencies:
  override:
  pre:
    - sudo apt-get update && sudo apt-get install -y --only-upgrade google-chrome-stable
    - npm install -g bower
  post:
    - bower install

```

deepflame:
@martndemus you could also change your build environment from Ubuntu 12.04 (Precise) to Ubuntu 14.04 (Trusty). That comes with Chrome 50 out of the box.

You may do the change here:
https://circleci.com/gh/DockYard/ember-composable-helpers/edit#build-environment

martndemus:
@nicomihalich Is this good to go?

nicomihalich:
@martndemus Should be.  @poteto Can you :eyes: one more time? Thanks!

package.json
```diff
@@ -27,26 +27,28 @@
...
     "ember-cli-htmlbars-inline-precompile": "^0.3.1",
     "ember-cli-inject-live-reload": "^1.4.0",
     "ember-cli-jshint": "^1.0.0",
-    "ember-cli-qunit": "^1.4.0",
+    "ember-cli-qunit": "^2.0.0",
     "ember-cli-release": "^0.2.9",
     "ember-cli-sri": "^2.1.0",
+    "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
+    "ember-data": "^2.7.0",
```
poteto: Delete ember data, this add-on doesn't use it

circle.yml
```diff
@@ -2,13 +2,8 @@ machine:
   node:
     version: "stable"
 dependencies:
-  override:
-    - curl -L -o google-chrome.deb https://s3.amazonaws.com/circle-downloads/google-chrome-stable_current_amd64_47.0.2526.73-1.deb
-    - sudo dpkg -i google-chrome.deb
-    - sudo sed -i 's|HERE/chrome\"|HERE/chrome\" --disable-setuid-sandbox|g' /opt/google/chrome/google-chrome
-    - rm google-chrome.deb
   pre:
+    - sudo apt-get update && sudo apt-get install -y --only-upgrade google-chrome-stable
```
martndemus: According to @poteto, this line is redundant.

PR #211, opened 2016-08-19
html-safe helper

offirgolan:
Add support for Ember.String.htmlSafe. 

Usage: 

```hbs
<button class={{html-safe (join '  ' myClasses)}}>Custom Button</button>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Capitalizes a string using `Ember.String.underscore`.
 **[⬆️ back to top](#available-helpers)**
 
 #### `html-safe`
-Mark a string as safe for unescaped output with Ember templates using `Ember.String.htmlSafe`
+Mark a string as safe for unescaped output with Ember templates using `Ember.String.htmlSafe`.
 
 ```hbs
 {{html-safe "<div>someString</div>"}}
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -266,6 +267,15 @@ Capitalizes a string using `Ember.String.underscore`.
 ```
 **[⬆️ back to top](#available-helpers)**
 
+#### `html-safe`
+Mark a string as safe for unescaped output with Ember templates using `Ember.String.htmlSafe`
```
poteto: Tiny formatting issue: needs a `.` at the end

app/helpers/html-safe.js
```diff
@@ -0,0 +1 @@
+export { default, invoke } from 'ember-composable-helpers/helpers/html-safe';
```
poteto: Should be `export { default, htmlSafe } from ...`

offirgolan: Oh whoa good catch! 😅

offirgolan:
@poteto fixes have been added.

poteto:
Thanks! @homu r+

PR #213, opened 2016-08-23
Pass parameters to `optional` handlers

TimGThomas:

The existing `optional` helper doesn't pass any specified parameters to the action handler. This PR adds this feature.

### DIFF
--- a/addon/helpers/optional.js
+++ b/addon/helpers/optional.js
@@ -1,8 +1,8 @@
 import { helper } from 'ember-helper';
 
-export function optional([action]) {
+export function optional([action, ...args]) {
   if (typeof action === 'function') {
-    return action;
+    return action.bind(this, ...args);
   }
 
   return (i) => i;
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Toggles a boolean value.
 Allows for the passed in action to not exist.
 
 ```hbs
-<button {{action (optional handleClick)}}>Click Me</button>
+<button {{action (optional handleClick "foo")}}>Click Me</button>
 ```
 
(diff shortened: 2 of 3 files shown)


addon/helpers/optional.js
```diff
@@ -1,8 +1,8 @@
 import { helper } from 'ember-helper';
 
-export function optional([action]) {
+export function optional([action, ...args]) {
   if (typeof action === 'function') {
-    return action;
+    return action.bind(this, ...args);
```
martndemus: Please avoid binding `this`, because that will screw up the context of (closure) actions.

TimGThomas: Makes sense. Should that be `null` instead, or do we have access here to the calling context?

martndemus: Certainly not `null` as that would set the context to `window`. I don't actually know what it should be. You should add a test using `this` in the action's context and then figure out what works.

TimGThomas: Yeah; I'm not sure what it should be, either. Honestly, I'm having a hard time even finding an example of a closure action with a custom context.

Regardless of the test case, I think this could be solved by returning a closure of the action and doing away with any `bind()` silliness altogether; would you agree?

```js
if (typeof action === 'function') {
  return () => action(...args);
}
```

martndemus: Yeah, that might be it.
But you might need to also factor in the arguments passed into the anonymous function you're returning from the helper.

TimGThomas: Agreed. Gonna take a deeper look this evening and hopefully have some more test cases soon!

martndemus:
Closing this due to inactivity.

PR #227, opened 2016-09-17
Make compatible with Ember 2.9.0-beta.2+.

rwjblue:
* Avoid mutating params/hash in helpers.
* Prevent errors when `ACTION` isn't needed.


poteto:
😍  Thank you!

PR #236, opened 2016-10-31
Use Ember.isEqual

### LINKED ISSUE #235: Use `Ember.isEqual` to check equality
[`Ember.isEqual`](http://emberjs.com/api/#method_isEqual) is a utility that was introduced in Ember 2.1.0 to make it possible to compare complex objects to other complex objects or primitive values. 

`ember-microstates` wraps primitive values in an object which breaks equality comparison. You can read more about this in [String microstate instance is not strictly equal to it's primitive value](https://github.com/cowboyd/ember-microstates/issues/61) issue. 

> What this feature means for your product?

`Ember.isEqual` will make it possible for us to add compatibility to `ember-composable-helpers`.

> What will it allow you to do that you can't do today? 

Currently, passing a microstate into a helper like `filter-by` will produce wrong results because `filter-by` uses strict comparison(`===`). 

If you take an example like this:

```hbs
{{let items=(array (hash type="animal" name="dog") (hash type="thing" name="boat))}}
{{let type=(String "animal)}}

{{#each (filter-by "type" type items) as |item|}}
  {{item.name}}
{{/each}}
```

This example will not filter the results because `(String "animal")` produces an object similar to `new String('animal')`, which is loosely equal to `"animal"` but not sticktly equal as `filter-by` expects.

> How will it make current work-arounds straightforward? 

Current work around is to use `value-of` helper that'll call `.valueOf()` on the passed in argument and return the primitive value.

```diff
{{let items=(array (hash type="animal" name="dog") (hash type="thing" name="boat))}}
{{let type=(String "animal)}}

- {{#each (filter-by "type" type items) as |item|}}
+ {{#each (filter-by "type" (value-of type) items) as |item|}}
  {{item.name}}
{{/each}}
```

If `ember-composable-helpers` uses `Ember.isEqual`, then we'll be able to make `value-of` uneccessary when used with `ember-composable-helpers`.

## Version
Ember 2.1.0+

## Test Case
You can see a [passing test case with `value-of`](https://github.com/cowboyd/ember-microstates/commit/00fe94b682b6f1836aa6727cb6cac83cf183b769) in the branch of `ember-microstates`.

## Expected Behavior
I expect `ember-composable-helpers` to know how to properly evaluate equality of objects that implement `isEqual` interface.

## Actual Behavior
`ember-composable-helpers` does not support objects that implement `isEqual` interface.

## Comments

1. There [is an issue](https://github.com/cowboyd/ember-microstates/issues/62) in `ember-microstates` library to add `isEqual` interface to all objects created by the library.
2. If this proposal is accepted, I'll provide a PR that'll replace all instances of `===` and `!==` with `isEqual` and `!isEqual`

taras:
This PR modifies `utils/isEqual` to use `Ember.isEqual`. `Ember.isEqual` only checks if the first argument implements `isEqual` interface, so I have to check twice: first for second argument against first, followed by first argument against second.

Closes #235

### DIFF
--- a/addon/helpers/filter-by.js
+++ b/addon/helpers/filter-by.js
@@ -6,6 +6,7 @@ import get from 'ember-metal/get';
 import observer from 'ember-metal/observer';
 import set from 'ember-metal/set';
 import { isEmpty, isPresent } from 'ember-utils';
+import isEqual from '../utils/is-equal';
 
 const { defineProperty } = Ember;
 
@@ -38,7 +39,7 @@ export default Helper.extend({
       if (typeof value === 'function') {
         filterFn = (item) => value(get(item, byPath));
       } else {
-        filterFn = (item) => get(item, byPath) === value;
+        filterFn = (item) => isEqual(get(item, byPath), value);
       }
     } else {
       filterFn = (item) => isPresent(get(item, byPath));
--- a/addon/helpers/reject-by.js
+++ b/addon/helpers/reject-by.js
@@ -6,6 +6,7 @@ import get from 'ember-metal/get';
 import observer from 'ember-metal/observer';
 import set from 'ember-metal/set';
 import { isEmpty, isPresent } from 'ember-utils';
+import isEqual from '../utils/is-equal';
 
 const { defineProperty } = Ember;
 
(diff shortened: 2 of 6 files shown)


martndemus:
Thanks!

taras:
Awesome. Thank you!

martndemus:
If there are more `===` in the codebase I'm happy to merge PR's to have those replaced by isEqual.

taras:
I looked for `===` and `!==` - only changed once that made sense.

martndemus:
:+1:

xomaczar:
npm publish?

PR #240, opened 2016-11-15
Allow inc/dec to take float

ming-codes:
The goal of PR #174 was to parse string values to number, but it only handles integer case. This PR is to enhance it by allowing any number, not just integers.


ming-codes:
@poteto?

Issue #247, opened 2017-01-06
isPromise does not return true for native promises


Asherlc:
## Version
1.0.0

## Test Case
isPromise(Promise.resolve());

## Steps to reproduce
Not sure there's anything more to this -- maybe misunderstanding?

## Expected Behavior
Return true

## Actual Behavior
Returns false

`#finally` isn't part of the Promise spec, but it's in `RSVP.Promise` and other libraries, so that condition should be removed.

poteto:
`isPromise` is private API for this addon

Asherlc:
@poteto if you use the `pipe` helper with a native Promise-returning action, it won't wait to fire the next action. Maybe I wrote this up wrong?

poteto:
Ah I see, that makes sense now. Yes, in that case would you like to open a PR to remove the `finally` check?

Asherlc:
@poteto sure thing, check out https://github.com/DockYard/ember-composable-helpers/pull/248

PR #248, opened 2017-01-09
Update is-promise.js

Asherlc:
Remove checking for `finally` on promises, since that's not in the promise spec

Fixes https://github.com/DockYard/ember-composable-helpers/issues/247


poteto:
Thanks!

PR #252, opened 2017-01-28
Upgrade to ember-cli/ember @2.11.0

xomaczar:


addon/-private/create-needle-haystack-helper.js
```diff
@@ -8,6 +8,7 @@ import set from 'ember-metal/set';
 const {
   isEmpty
 } = Ember;
+const K = () => { };
```
xomaczar: see line line 22

PR #257, opened 2017-03-06
Add failing test to queue related to conditional actions

cbou:
Queue helper does not support conditional actions.

I'm not sure if it's a bug with EmberJS or ember-composable-helpers...


snewcomer:
Closing for now.  If someone needs this feature, would ❤️ a PR!

PR #259, opened 2017-03-28
enhance slice helper to also support passing only 2 arguments

badazz91:
In order to support native ``` slice ``` behaviour like this:

```
a = [1,2,3,4]
a.slice(-1) -> [4]
```
Until now the slice helper expected always ```start```, ```stop``` and ```array``` being passed to the helper like this ```{{slice 0 2 array}}```.

Now it is possible to use the slice helper also like this:
```{{slice -1 array}}```


badazz91:
Unfortunetly there was already some failing tests when I checked out the current master.

PR #273, opened 2017-06-06
Update ember-cli to 2.13

cibernox:

Update the project to the 2.13 family.

Other changes:
- Replace `jshint` by `eslint` and use pull ember-suave from the eslint package.
- Fix typo in package.json
- Migrate to `ember-native-dom-helpers` using a codemod. Still I haven't remove jQuery.

Notes:
- There is two failing tests, but those failures were also in master. Apparently you no longer can observer changes to the action  being passed to a helper. 
E.g: `{{#each (map (action (if flag "cb1" "cb)) array) as |foo|}}`
Those tests pass in 2.8 but fail in 2.11+. I suspect that glimmer has changed this.


.eslintrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  root: true,
+  parserOptions: {
+    ecmaVersion: 2017,
+    sourceType: 'module'
+  },
+  extends: 'eslint:recommended',
```
brzpegasus: You added `eslint-plugin-ember-suave` to the project but are not using it.

brzpegasus:
> There is two failing tests, but those failures were also in master. Apparently you no longer can observer changes to the action being passed to a helper.
E.g: {{#each (map (action (if flag "cb1" "cb)) array) as |foo|}}
Those tests pass in 2.8 but fail in 2.11+. I suspect that glimmer has changed this.

Likely related: https://github.com/emberjs/ember.js/issues/14748

cibernox:
@brzpegasus According to that issue, the fact that when the inputs of an action change the closure action updates was not really a public API. It was incidental and it's not supported. Seems that they don't plan to support it.

I vote to delete those two tests and the observer that tries to check for changes on the action, since its not really doing anything.

brzpegasus:
> I vote to delete those two tests and the observer that tries to check for changes on the action, since its not really doing anything.

This makes sense to me.

package.json
```diff
@@ -16,43 +16,44 @@
   "bugs": "https://github.com/DockYard/ember-composable-helpers/issues",
   "homepage": "https://github.com/DockYard/ember-composable-helpers",
   "engines": {
-    "node": ">= 0.10.0"
+    "node": ">= 4"
   },
-  "author": [
+  "authors": [
```
brzpegasus: @cibernox Is this the typo you mentioned in the PR description? NPM's `package.json` only supports a single [author](https://docs.npmjs.com/files/package.json#people-fields-author-contributors) (but can have multiple _collaborators_). You would have to add an `AUTHORS.md` file at the root of the project. An alternative would be to set the author as `DockYard <oss@dockyard.com>` and list these existing entries under `collaborators`. That could probably be separated from this PR though.

cibernox: I'll do it after then.

PR #276, opened 2017-06-30
group-by helper no longer silently drops grouped items 

sirvine:

(Related to #234.)

Currently, `group-by` resets the grouped items on each pass through the iterator when `groupName` is complex string that includes punctuation such as periods or commas. The culprit was`get(groups, groupName)`, which fails to match the correct `groupName` key.

### DIFF
--- a/addon/helpers/group-by.js
+++ b/addon/helpers/group-by.js
@@ -18,13 +18,12 @@ const groupFunction = function() {
 
   array.forEach((item) => {
     let groupName = get(item, byPath);
-    let group = get(groups, groupName);
+    let group = groups[`${groupName}`];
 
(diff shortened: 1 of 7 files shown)


addon/helpers/group-by.js
```diff
@@ -18,13 +18,12 @@ const groupFunction = function() {
 
   array.forEach((item) => {
     let groupName = get(item, byPath);
-    let group = get(groups, groupName);
+    let group = groups[`${groupName}`];
```
adamjmcgrath: `groups` could be a POJO in this case

snewcomer:
These helpers are significantly refactored.  Let us know if you are still having these problems!

PR #277, opened 2017-07-17
Updating without helper to work with ember data arrays. Fixes #268.

### LINKED ISSUE #268: `without` does not treat `hasMany` relationship data like arrays
## Version
2.0.1

## Test Case
Assume there is a model `option` with just a `name: attr('string')` and another model with a relationship `options: hasMany('option')`.

## Steps to "reproduce"
Load all options such that `{{map-by 'name' allOptions}}` is `'a','b','c'` and add one relation to an item of the latter type such that `{{map-by 'name' item.options}}` is `'b'`.

## Expected Behavior
`{{map-by 'name' (without item.options allOptions)}}` should be `'a','c'`.

## Actual Behavior
`{{map-by 'name' (without item.options allOptions)}}` is still `'a','b','c'`. 

(`{{without (map-by 'name' item.options) (map-by 'name' allOptions)}}` actually is `'a','c'`.)

## Probable Cause
[The line `if (typeOf(needle) === 'array' && get(needle, 'length')) {`](https://github.com/DockYard/ember-composable-helpers/blob/3255f87bbe46b343c29b60628a920c1037833c91/addon/helpers/without.js#L19), because `typeOf(item.options)` is 'instance'. It does have a `length`, though.

gmurphey:
Closes #268.

Ember Data collections return `instance` when run through `Ember.typeOf`. This PR changes the array check of `needle` to use `isEmberArray`, which also detects Ember Data collections.

### DIFF
--- a/addon/helpers/without.js
+++ b/addon/helpers/without.js
@@ -1,6 +1,5 @@
 import { A as emberArray, isArray as isEmberArray } from '@ember/array';
 import { get } from '@ember/object';
-import { typeOf } from '@ember/utils';
 import createNeedleHaystackHelper from '../-private/create-needle-haystack-helper';
 import includes from '../utils/includes';
 
@@ -13,7 +12,7 @@ export function without(needle, haystack) {
     return false;
   }
 
-  if (typeOf(needle) === 'array' && get(needle, 'length')) {
+  if (isEmberArray(needle) && get(needle, 'length')) {
     return haystack.reduce((acc, val) => contains(val, needle) ? acc : acc.concat(val), []);
   }
 
--- a/tests/integration/helpers/without-test.js
+++ b/tests/integration/helpers/without-test.js
@@ -72,3 +72,31 @@ test('it accepts array-like arrays', function(assert) {
 
   assert.equal(find('*').textContent.trim(), 'barbaz', 'should render remaining values');
 });
+
+test('it accepts an ember data array', function(assert) {
+  this.inject.service('store');
+
+  run(() => {
+    let person = this.get('store').createRecord('person', {
+      name: 'Adam'
+    });
+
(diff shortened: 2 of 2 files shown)


addon/helpers/without.js
```diff
@@ -16,7 +15,7 @@ export function without(needle, haystack) {
     return false;
   }
 
-  if (typeOf(needle) === 'array' && get(needle, 'length')) {
+  if (isEmberArray(needle) && get(needle, 'length')) {
```
sbatson5: Will this break for normal arrays? I.e. an array in a project where prototypes are not extended by default?

sbatson5: We could get around that with an or check
`if ((isEmberArray(needle) || typeOf(needle) === 'array') && get(needle, 'length'))`

gmurphey: `isEmberArray` is a shim for [`Ember.isArray`](https://www.emberjs.com/api/ember/2.14/namespaces/Ember/methods/isArray?anchor=isArray) which will return `true` for both native arrays and array-like objects.

Techn1x:
Can we get this merged in? It also fixes array attributes used in `ember-data-model-fragments`

Techn1x:
What's left to get this into ember-composable-helpers? I'm having to reference this commit as an npm package just for this PR

Subtletree:
Just hit this now. Merge would be great!

Techn1x:
@snewcomer @sbatson5 Is anything blocking this? Or does it just need rebasing?

snewcomer:
Looks good.  Will fix any tests after merge (just cause it has been outstanding for a while and avoid asking Garret to rebase)

https://www.emberjs.com/api/ember/3.7/functions/@ember%2Farray/isArray

PR #278, opened 2017-07-28
Use "New Module Imports"

Turbo87:
see https://github.com/ember-cli/ember-rfc176-data

/cc @cibernox

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,9 +8,13 @@ module.exports = {
     'eslint:recommended',
     'plugin:ember-suave/recommended'
   ],
+  plugins: [
+    'ember'
+  ],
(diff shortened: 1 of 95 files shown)


cibernox:
It looks good but there is still imports using the old shims we have to fix. Does the codemod support those?

Turbo87:
> It looks good but there is still imports using the old shims we have to fix.

looks like I missed those 🙈 

> Does the codemod support those?

no, but the `no-old-shims` ESLint rule does

Turbo87:
> It looks good but there is still imports using the old shims we have to fix

done :)

cibernox:
New shims all-the-things!

PR #279, opened 2017-08-21
Read options from parent or app

makepanic:

The PR makes ember-composable-helpers read its options from either `this.parent` or `this.app`.

The "issue" is that relying only on `this.app` makes it impossible to configure from within an engine.
The changed code is copied from [`ember-cli-babel`](https://github.com/babel/ember-cli-babel/blob/master/index.js#L137).

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,9 +18,9 @@ module.exports = {
     }
 
     this.app = app;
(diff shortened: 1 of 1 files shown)


poteto:
Thanks!

PR #281, opened 2017-09-27
Add sort and reverse-sort helpers

mwpastore:
Here is a first swag at adding `(sort)` and `(reverse-sort)` helpers. These simplified versions of the `(sort-by)` helper allow users to computed-sort arrays of values and not just arrays of objects. They use Ember.compare as the comparator and do not accept any additional arguments. 

* `(reverse-sort)`? `(sort-reverse)`? Something else?
* Should it just be one `(sort)` helper with a reverse boolean argument?


mwpastore:
Hrrm, I didn't touch `(previous)` or `(next)` and yet their tests are failing for some reason(s). Any suggestions?

PR #285, opened 2018-01-03
feat(tree-shaking): remove unused helpers from app tree

lennyburdette:
The `only` and `except` configuration options for tree-shaking affect the host app's vendor.js but not the application js.

We (Square) noticed this by installing the addon in an empty app and noticing that the application js file increased by 12kb regardless of how we configured tree-shaking.


snewcomer:
This was fixed by https://github.com/DockYard/ember-composable-helpers/pull/316

Thank you!

PR #287, opened 2018-01-18
remove broken exports

kellyselden:
These exports do not exist. It was detected with my tree shaking work.


snewcomer:
This seems to be fixed in another PR.  Thanks Kelly! #302

PR #292, opened 2018-03-02
Fixed range helper inclusive, when args are the same value

### LINKED ISSUE #291: Range Helper inclusive bug
## Version
2.10

## Test Case
https://github.com/DockYard/ember-composable-helpers/blob/master/tests/integration/helpers/range-test.js#L29-L36

## Steps to reproduce
Use the Range helper as inclusive with the same value for both args.
```handlebars
{{log (range 1 1 true)}}
```

## Expected Behavior
Result: [1]

## Actual Behavior
Result: []

-----

By default, the range helper is inclusive of the 1st value, and exclusive of the second.
(range 1 5) -> [1,2,3,4]
(range 1 2) -> [1]
(range 1 1) -> []

The inclusive flag should make the 2nd arg always included in the array
(range 1 5 true) -> [1,2,3,4,5]
(range 1 2 true) -> [1,2]
(range 1 1 true) -> [] // Should be [1]

Currently, (range 1 1) === (range 1 1 true), which doesn't seem right

-----

Fix is pretty straightforward, just need the following line added to https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/range.js#L22
```
if(min === max && isInclusive) return [max];
```

Techn1x:
Closes #291

### DIFF
--- a/addon/helpers/range.js
+++ b/addon/helpers/range.js
@@ -20,6 +20,10 @@ export function range([min, max, isInclusive]) {
     }
   }
 
+  if (min === max && isInclusive) {
+    numbers.push(max);
+  }
+
   return numbers;
 }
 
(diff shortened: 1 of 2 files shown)


Techn1x:
The failing tests don't seem to be to do with my changes ¯\_(ツ)_/¯

PR #288, opened 2018-01-27
Updating ember-cli and fixing tests

gmurphey:
## Changes proposed in this pull request

Noticed that tests were failing after rebasing #277. This PR updates to the latest ember-cli and fixes failing tests.

PR #296, opened 2018-04-05
pipe/queue exception handling

dadleyy:

The `pipe` and `queue` helper methods are great for composing action chains and keeping responsibility of any given piece of action logic autonomous. This is particularly useful for library components that allow consumers to "tap into" the action pipeline. Consider the following piece of generic/library template:

```hbs
{{!-- my-dropdown/contextual-button.hbs --}}
<button onclick={{pipe (action @save) (action 'save') (action dropdown.close)}}>
 Save
</button>
```

In this example, the component "relinquishes" control at the first stage of the pipe to the consumer where they are allowed to provide a `save` method that will be called by the component. The consumer would look something like this:

```hbs
{{!-- consumer.hbs --}}
{{#my-dropdown as |dropdown|}}
  {{dropdown.save-button save=(action 'save')}}
{{/my-dropdown}}
```

If the consumer's `save` action were to raise an exception, the `my-dropdown/contextual-button` component would not be able to handle it; the `pipe` helper stops execution after the first exception. This would be unfortunate if the `my-dropdown` component wanted to close the dropdown regardless of what happened in it's consumer's `save` method.

This change introduces the `catch` and `finally` named arguments to the pipeline helpers to allow components to execute logic to handle exceptions raised during the pipe or to have logic executed regardless of successful completion:

```hbs
{{!-- my-dropdown/contextual-button.hbs --}}
<button onclick={{pipe (action @save) (action 'save') finally=(action dropdown.close)}}>
 Save
</button>
```

#### Breaking changes

The `pipe` &amp; `queue` methods that are exported from the addon now _always_ return a promise.

#### Note

I branched this off [`#288`]; it looks like it has the CI environment shaped up nicely. I will rebase this branch with master once that lands. This does mean that the diff of this PR includes @gmurphey's changes. The only changes this PR is concerned with is [`dc669bf`].

[`#288`]: https://github.com/DockYard/ember-composable-helpers/pull/288
[`dc669bf`]: https://github.com/DockYard/ember-composable-helpers/pull/296/commits/dc669bf66535b84a97a318931464f3e403592b07


seanCodes:
@dadleyy @cibernox @snewcomer Wondering if there was a reason this was closed, or if this feature is something you'd still consider accepting? I’d be happy to carry the torch and continue the work if so.

PR #301, opened 2018-09-01
Add query by helper

### LINKED ISSUE #298: [FEATURE REQUEST] Add 'includes' helper for checking substring in a string
A `includes` helper for checking whether a [string contains a substring](https://stackoverflow.com/questions/1789945/how-to-check-whether-a-string-contains-a-substring-in-javascript) could be handy.
Example -

```handlebars
{{#if (includes "apples" "apples and oranges")}}
  Apple and Orange!
{{/if}}
```

I was hoping the `contains` helper does this but looks like it only works on arrays.

brunoocasali:
- I've created a simple function to query array items from a matched query. This package has contains but is only for exactly contents. 

Any suggestion will be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ For help upgrading between major versions, check out the [upgrading documentatio
   + [`has-next`](#has-next)
   + [`previous`](#previous)
   + [`has-previous`](#has-previous)
+  + [`query-by`](#query-by)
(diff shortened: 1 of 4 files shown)


snewcomer:
I think this type of logic may be best suited for individual needs!

PR #302, opened 2018-09-05
Remove reexports of nonexistent names

ef4:
These things are all reexports of names that don't actually exist. If anybody tries to use these, they get an error. But almost certainly nobody tries to use these, because they shouldn't anyway. They should import the originals directly out of the addon's namespace, not the app's namespace.

I noticed this issue because I'm experimenting with more automatic dependency traversal, and these bad links generate warnings.


poteto:
Thanks!

PR #312, opened 2019-02-18
Add yarn

snewcomer:
Try to get circle to pass.  Errors using npm and haven't found a way around.


snewcomer:
This seems to fix our issues with tests.

PR #316, opened 2019-04-18
fix unused helper stripping

### LINKED ISSUE #308: `only` and `except` leave broken module references and unnecessary code in app tree
The `only` and `except` options remove the implementations of unused helpers from the *addon* tree, but they don't remove the corresponding reexport modules from the *app* tree.

This leaves unnecessary and broken modules in the resulting application.

The same problem exists in the `addon/index.js` file, which will try to import every helper regardless of whether it has been stripped out. This makes it impossible to import the remaining helpers via that module.

(I noticed because this breaks [embroider](https://github.com/embroider-build/embroider) and will break any system that tries to verify imports at build time.)

ef4:
This fixes #308.

As there is no existing test infrastructure for running builds with different build-time options, there was not a good place to put a test so I didn't make one.

The changes here fix both parts of the problem reported in #308.

 - in the addon tree, we need to rewrite the `index.js` file so it stops trying to import helpers that have already been removed.
 - in the app tree, we need to remove the files that reexport removed helpers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ var Funnel = require('broccoli-funnel');
 var path = require('path');
 var intersection = require('./lib/intersection');
 var difference = require('./lib/difference');
+var StripBadReexports = require('./lib/strip-bad-reexports');
 
 module.exports = {
   name: require('./package').name,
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -24,11 +25,15 @@ module.exports = {
...
 
-  treeForAddon: function() {
-    // see: https://github.com/ember-cli/ember-cli/issues/4463
-    var tree = this._super.treeForAddon.apply(this, arguments);
-    var moduleRegexp = '(^modules\/)?' + this.name + '\/helpers\/';
-    return this.filterHelpers(tree, new RegExp(moduleRegexp, 'i'));
+  treeForAddon: function(tree) {
+    tree = this.filterHelpers(tree, new RegExp(/^helpers\//, 'i'));
+    tree = new StripBadReexports(tree, [`index.js`]);
+    return this._super.treeForAddon.call(this, tree);
```
ef4: Life is easier if you process the tree _before_ it has run through `_super.treeForAddon`. For one thing, you no longer need to worry about whether your current version of ember-cli includes the `modules` subdir (which is why the original RegExp was more complex). For another, we want our StripBadReexports plugin to process the import statements before they have been compiled into AMD by ember-cli-babel.

index.js
```diff
@@ -38,7 +43,7 @@ module.exports = {
 
     // exit early if no opts defined
     if ((!whitelist || whitelist.length === 0) && (!blacklist || blacklist.length === 0)) {
-      return new Funnel(tree);
+      return tree;
```
ef4: `new Funnel(tree)` is a no-op.

PR #319, opened 2019-05-09
refactor(filter-by): PoC fake sync observer

### LINKED ISSUE #318: `sort-by` doesn't seem to work with the latest Octane blueprint
The following template shows [`sort-by`](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/sort-by.js) working in Ember `3.9.1` and `3.10.0-beta.5` but not in an app using the latest Octane blueprint (`ember new octane-app -b @ember/octane-app-blueprint`):

**application.hbs**
```hbs
{{#let (array (hash age=1) (hash age=5) (hash age=2)) as |rows|}}
  <h4>Raw Rows:</h4>
  {{#each rows as |row|}}
    [row: {{row.age}}]
  {{/each}}

  <hr>

  <h4>Sorted Rows:</h4>
  {{#each (sort-by "age" rows) as |row|}}
    [sorted: {{row.age}}]
  {{/each}}
{{/let}}
```

Screen Shot 2019-05-02 at 09 42 22

Screen Shot 2019-05-02 at 09 35 55

~~possibly related to https://github.com/DockYard/ember-composable-helpers/issues/289~~

buschtoens:
Closes #318.
/cc @snewcomer @GavinJoyce

I'm currently evaluating different possible solutions.

This one is very hacky and I'm not suggesting that we do something like this, _but_: I replaced the `byPathDidChange` observer in `filter-by` with a "`syncObserver`", that installs setters on the dependent keys that synchronously calls the observer callback.

Of course this is a mega hack and is also no full sync observer polyfill, but it uncovered another problem. If I understood the conversation in #318 correctly, one suggestion was to make `byPathDidChange` sync / replace the observer with something else, but to keep `contentDidChange` as an async observer.

This does not work. As you can see in the tests, I needed to add `await settled()` for the tests where the functionality of `contentDidChange` is tested. This is because the observer, well, the observer is _async_ and AFAIK only gets executed in the next run loop.

### DIFF
--- a/addon/helpers/filter-by.js
+++ b/addon/helpers/filter-by.js
@@ -7,6 +7,7 @@ import { observer } from '@ember/object';
 import { set } from '@ember/object';
 import { isEmpty, isPresent } from '@ember/utils';
 import isEqual from '../utils/is-equal';
+import syncObserver from '../utils/sync-observer';
 
 export default Helper.extend({
   compute([byPath, value, array]) {
@@ -22,7 +23,7 @@ export default Helper.extend({
     return get(this, 'content');
   },
 
-  byPathDidChange: observer('byPath', 'value', function() {
+  byPathDidChange: syncObserver('byPath', 'value', function() {
     let byPath = get(this, 'byPath');
     let value = get(this, 'value');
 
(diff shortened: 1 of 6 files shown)


buschtoens:
Indeed, instead of `await settled()` I can also make the tests pass by forcing a second runloop via `run(() => {})`.

snewcomer:
@buschtoens What do you think the status of this PR is (nice work btw)?  Been reading up on this issue and it seems like a future PR will be necessary?

GavinJoyce:
FYI I asked [a question](https://github.com/emberjs/rfcs/pull/494#issuecomment-497603878) about what needs to be done to migrate this addon to a world without observers, here's [pzuraq's response](https://github.com/emberjs/rfcs/pull/494#issuecomment-497796611):

Screen Shot 2019-06-05 at 15 36 20

snewcomer:
👍 What is autotracking as it relates to Ember/Glimmer?  I hadn't seen that word thrown around until seeing that response.

GavinJoyce:
we should be able to close this now that we have https://github.com/DockYard/ember-composable-helpers/pull/321

Issue #224, opened 2016-09-07
Helper performance improvements


aaronbhansen:
I was stepping through the helpers yesterday and noticed many of them doing extra computations that could be avoided. The use of observers aren't needed in many of the helpers today. For this example, lets take the filter-by helper.

**One method**
compute

**Two observers**
byPathDidchange, contentDidChange

Calling the helper, currently the following execution chain happens.

```
(filter-by "property" "value" array)
```

Compute is called on the helper
- on line 20 byPath is set.
- Since observers are synchronous, byPathDidChange is executed. 1st time.
- on line 21 value is set.
- byPathDidChange is executed again for 2nd property change. 2nd time called.

A property is defined twice that only needs to be defined once. 

If at any point any of the parameters passed in is changed, the helper will be called again and compute is executed. 

By its nature, the observer byPathDidChange is never called because the parameters its observing can't change unless compute is called. Compute is called anytime a parameter changes, so no need to observe on those properties. 

contentDidChange is still needed, but any other observers could be removed. 

Proposed solution. 
1. Remove as many observers as possible. They aren't needed and are extra work being done
2. Change byPathDidChange to a method -> initializeProperty and call that from compute.
3. Don't set any extra properties to helper than needed. The only real property needing to be set is array, the others can be passed in an used within the initializeProperty scope. 

Possible solution

```
import Ember from 'ember';
import { isEmberArray } from 'ember-array/utils';
import { filter } from 'ember-computed';
import Helper from 'ember-helper';
import get from 'ember-metal/get';
import observer from 'ember-metal/observer';
import set from 'ember-metal/set';
import { isEmpty, isPresent } from 'ember-utils';

const { defineProperty } = Ember;

export default Helper.extend({
  compute([byPath, value, array]) {
    if (!isEmberArray(array) && isEmberArray(value)) {
      array = value;
      value = undefined;
    }

    set(this, 'array', array);
    this.initializeProperty(byValue, value);

    return get(this, 'content');
  },

  initializeProperty(byPath, value) {
    if (isEmpty(byPath)) {
      defineProperty(this, 'content', []);
      return;
    }

    let filterFn;

    if (isPresent(value)) {
      if (typeof value === 'function') {
        filterFn = (item) => value(get(item, byPath));
      } else {
        filterFn = (item) => get(item, byPath) === value;
      }
    } else {
      filterFn = (item) => isPresent(get(item, byPath));
    }

[...]

PR #321, opened 2019-09-26
[Breaking Change] Refactor helpers for Octane (Ember 3.13+)

GavinJoyce:
/cc @buschtoens @snewcomer 

closes https://github.com/DockYard/ember-composable-helpers/issues/224

Now that [arguments to helpers are auto-tracked](https://github.com/emberjs/rfcs/pull/494#issuecomment-497603878), we can refactor the helpers to remove their use of observers and computed properties. This should result in a significant performance improvement as synchronous observers as relatively expensive.

**Breaking changes:**

 * This addon now depends on Ember 3.13+
 * removed `array` helper as it is [included in Ember since 3.8](https://blog.emberjs.com/2019/02/27/ember-3-8-released.html)

TODO:

 * [x] refactor all helpers (except `sort-by`, we can do that in a future PR: https://github.com/DockYard/ember-composable-helpers/issues/327)
 * [x] remove param munging duplication
 * [x] try this on Intercom: 🍏 
 * [x] remove `array` helper (as it's [included in ember](https://api.emberjs.com/ember/3.13/classes/Ember.Templates.helpers/methods/array?anchor=array) since 3.8)
 * [x] rebase once https://github.com/DockYard/ember-composable-helpers/pull/323 lands
 * [x] Add a compatibility note to readme

related to https://github.com/DockYard/ember-composable-helpers/issues/318#issuecomment-497635938


snewcomer:
@GavinJoyce 💯 I agree.  Mind merging in node 8 commit from master?  Also, trying to understand this better.  

1. Will this require a major version bump?  Only supporting v3.0 for apps > 3.13?
2. He says observers can be replaced by "autotracking".  Is this something explicit we also need to integrate or implicit to glimmer templates?  I wasn't exactly sure what "autotracking" meant...

GavinJoyce:
> Will this require a major version bump? Only supporting v3.0 for apps > 3.13?

I'm not sure yet, it may be easy to conditionally install the observers for < 3.13. If it's tricky, a major version bump might be pragmatic

> He says observers can be replaced by "autotracking". Is this something explicit we also need to integrate or implicit to glimmer templates? I wasn't exactly sure what "autotracking" meant...

I believe that arguments to helpers are now autotracked, meaning that they will invalidate when they change and cause helpers to re-render. I'm going to go through the helpers now and try to remove all observers - we'll see if we can keep the tests passing

> Mind merging in node 8 commit from master?

I've rebased. We're still seeing a single test failure, unrelated to my changes:

Screenshot 2019-10-10 at 09 34 33

GavinJoyce:
> Another plus for a major version bump is that we can likely reimplement these helpers in modern ember, removing the need for using internal computed properties.

Eg:

Screenshot 2019-10-21 at 11 46 25

I think it's worth rebuilding all these helpers for Octane and releasing as a new major version.

GavinJoyce:
This is slightly simpler now 😃:

Screenshot 2019-10-21 at 12 11 58

addon/helpers/has-next.js
```diff
@@ -1,13 +1,27 @@
...
+  let currentValue = params[0];
+
+  let array;
+  let useDeepEqual = false;
+  if (params.length === 2) {
+    array = params[1];
+  } else {
+    useDeepEqual = params[1];
+    array = params[2];
+  }
```
GavinJoyce: `next`, `previous`, `has-next` and `has-previous` seem to have a slightly weird argument ordering where the optional `useDeepEqual` can be in the middle of two required arguments `currentValue` and `array`.

I wonder if we can reorder them as part of the breaking change?

snewcomer: that sounds fine to me.  I'll make sure we document this!  Perhaps `warn` if that arg is still a boolean?

addon/helpers/join.js
```diff
@@ -1,20 +1,13 @@
...
-import { observer } from '@ember/object';
-import { set } from '@ember/object';
 
-export default Helper.extend({
-  compute([separator, array]) {
-    if (isEmberArray(separator)) {
-      array = separator;
-      separator = ',';
-    }
+function join([separator, array]) {
```
GavinJoyce: Perhaps we should also switch the order of these so that the optional argument goes at the end? 

```hbs
{{join this.items}}
{{join this.items "-"}}
```

If we like, we can emit a deprecation warning if a string is passed as the first argument

GavinJoyce: It might be worth considering the API for all these helpers, eg:

perhaps:

`(map-by "fullName" users)`

should be?

`(map-by users "fullName)`

addon/helpers/object-at.js
```diff
@@ -15,22 +11,6 @@ export function objectAt(index, array) {
   return emberArray(array).objectAt(index);
```
GavinJoyce: for a future PR: we can likely improve performance by, instead of converting to an `emberArray`, checking if the array is already and `emberArray` and using `objectAt` or, if it's a native array, using `array[index]`

addon/helpers/join.js
GavinJoyce: I now see that the pattern of providing the array as the last argument is followed throughout the addon. I think we should leave the APIs as is in this PR and have a discussion on some future API in an issue. I'll open one soon

addon/helpers/union.js
```diff
@@ -1,4 +1,7 @@
-import { union } from '@ember/object/computed';
-import createMultiArrayHelper from '../-private/create-multi-array-helper';
+import { helper } from '@ember/component/helper';
 
-export default createMultiArrayHelper(union);
+export function union([...arrays]) {
+  return [...new Set([].concat(...arrays))];
```
GavinJoyce: https://stackoverflow.com/a/27664971/10879402

addon/helpers/join.js
GavinJoyce: > For all non-unary helpers, the subject of the helper function will always be the last argument. This way the arguments are better readable if you compose together multiple helpers

ah

addon/helpers/has-next.js
GavinJoyce: I've decided to leave the API as is, it seems the defining the `array` in the last position is by design

addon/helpers/union.js
makepanic: I'm not sure what IE11 compat should be but without a polyfill, `new Set(iterable)` isn't supported: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility

GavinJoyce: I assumed that babel would allow us to use `Set`, I'm not positive though.

README.md
```diff
@@ -16,10 +16,18 @@ Composable helpers for Ember that enables more declarative templating. These hel
...
+**Ember 3.13+:**
+
 ```no-highlight
 ember install ember-composable-helpers
 ```
 
+**Ember 3.12 and below:**
+
+```no-highlight
+ember ember-composable-helpers@^2.3.1
```
snewcomer: 2.4.0 published!

GavinJoyce: 👍 updated

addon/helpers/join.js
chriskrycho: @GavinJoyce FWIW that pattern (stolen from FP languages) is *sooooo* 😍 once you get used to it. It's precisely what makes these *composable*: you can partially apply them (historically with `action`, now with `fn`) and have the next one *also* partially applied, and so on… and only supply the relevant data at the very end.

addon/helpers/union.js
chriskrycho: You *may* need to double-spread it? `new Set(...[].concat(...arrays))`.

addon/helpers/chunk.js
```diff
@@ -31,22 +26,6 @@ export function chunk(num, array) {
   }
 }
 
-export default Helper.extend({
```
snewcomer: So effectively, no more stateful helpers b/c of the way glimmer templates work?  Stateful was necessary b/c....updated values passed to helpers were not invalidated?

addon/helpers/union.js
GavinJoyce: I've tried both `Set` approaches in IE11. As @makepanic pointed out, they don't work without a polyfill. Instead, I'm now using the [`[].filter` approach](https://stackoverflow.com/a/14438954/10879402):

```js
export function union([...arrays]) {
  let items = [].concat(...arrays);

  return items.filter((value, index, array) => array.indexOf(value) === index);
}
```

addon/helpers/join.js
GavinJoyce: Thanks. Now that I see how it composes, I'm a fan.

GavinJoyce:
> Perhaps we can lets this simmer for a few days and see if we can get anybody else to test.

I'm going to run this in production in Intercom today.

addon/helpers/chunk.js
GavinJoyce: Yes, that was true for the arrays and objects. Helpers are now recomputed when they change

GavinJoyce:
We've been running this in production in Intercom without issue for the past few hours

leepfrog:
FWIW: the `group-by` helper seems faster in my application (didn't benchmark it) and works fine.  Also the `union` helper is working fine.  Thanks for the update!

GavinJoyce:
We've been running this in our Intercom production app for 2 says without issue

PR #323, opened 2019-10-10
Upgrade to Ember 3.13

GavinJoyce:
extracted from https://github.com/DockYard/ember-composable-helpers/pull/321

This has a single failure for `Unit | Helper | chunk: it handles arrays with computed property lengths`:

Screenshot 2019-10-10 at 09 57 16

It seems that this fails in 3.13:

```js
let ap = emberArrayProxy.extend({
      content: emberArray([1, 2, 3]),
      length: computed('content.[]', function() {
        return this.get('content.length');
      })
    }).create();
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "ember-qunit": "^4.1.2",
     "ember-resolver": "^5.0.1",
     "ember-sinon": "^3.0.0",
-    "ember-source": "~3.5.0",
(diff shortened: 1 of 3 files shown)


snewcomer:
@GavinJoyce Perfect 👍 Looks like that `length` computed key passed to the `ArrayProxy` in that chunk test is running over something.  If that is removed (don't think it is needed), we should be all good!

PR #325, opened 2019-10-18
Add noop helper

laurmurclar:
Adds a `noop` helper. 
This can be useful when chaining helpers that expect functions as input, where you don't always actually have a function to pass in, or want to optionally choose between calling a function and doing nothing (as given in the example). 

We found it useful when converting component javascript to template-only components as sometimes our click functions (in the javascript) would conditionally call a function argument. We found adding a noop helper to be more readable than something like `{{on 'click' (optional (if @shouldCallTheFunction @onClick))}}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ Toggles a boolean value.
 Returns an empty function.
 
 ```hbs
-<button {{action (if isDisabled (noop) handleClick)}}>Click Me</button>
+<div {{on "mouseenter" (if @isLoading (noop) @sendTrackingEvent))}}>Some content</div>
 ```
 
 **[⬆️ back to top](#available-helpers)**
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ For help upgrading between major versions, check out the [upgrading documentatio
   + [`pipe`](#pipe)
   + [`compute`](#compute)
   + [`toggle`](#toggle)
-  + [`no-op`](#no-op)
+  + [`noop`](#noop)
   + [`optional`](#optional)
   + [`queue`](#queue)
 * [Array](#array-helpers)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -191,6 +192,16 @@ Toggles a boolean value.
 
 **[⬆️ back to top](#available-helpers)**
 
+#### `no-op`
```
snewcomer: Thank you thank you!  What do you think about `noop` to simplify the process of reading/writing it?  (I don't know if `noop` is the correct vernacular though....)

laurmurclar: We had this debate internally (I originally wrote this for Intercom, the company I work for) and `no-op` won out 😂 I'm team `noop` too, but don't feel strongly either way 😄  People thought `no-op` read better but `noop` seemed a more common name for these things (Angular and jQuery both have functions called `noop` for example) 🤷‍♀ 

Let's go with `noop` 😄 I'll make that change now. 

Another thing about this section - We also learnt that making use of the `disabled` html attribute is a better (more semantic, accessible, concise) way to implement the behaviour in the example, but `noop` is still useful for other situations, or elements that `disabled` doesn't apply to. I think it's still probably the simplest/clearest way to illustrate the behaviour, so I'll leave it as is for now. Would you prefer a different example?

snewcomer: Either way, really appreciate this PR!  And great point about the example.  Does something like this work?

```js
<button {{on "click" (if isDisabled noop this.handleClick)}}>click</button>
```

Perhaps there is a better example that is a bit less contrived....

laurmurclar: That _nearly_ works - you have to surround the `noop` with parenthesis or it won't realise that its just a helper without params. If we're going with Octane syntax, it also makes more sense for `isDisabled` and `handleClick` to be args passed in from the parent (as this helper is useful for making components template-only). 

So that would give us:
```html
<button {{on "click" (if @isDisabled (noop) @handleClick)}}>click</button>
```

If we want to avoid using the `button` (because of accessibility concerns) then maybe something like:
```html
<div {{on "mouseenter" (if @isLoading (noop) @sendTrackingEvent)}}>Some content</div>
```
Any preference?

snewcomer: I like the second one!  great idea

laurmurclar: Great! Pushed that change. Can squash the changes into one commit if you'd like?

laurmurclar:
Thanks @snewcomer ! 😄

laurmurclar:
And thank you for being so welcoming/the pleasant contributing experience!! 💃

PR #329, opened 2019-10-31
Handle null array in map-by

jrjohnson:
This avoids "array is null" when you use {{map-by}} with a null array.

This was broken by #323 when it was updated to `map` on the array directly instead of using the Ember computed `mapBy`.


jrjohnson:
Looks like there are a few more of these scattered around, adding tests and finding them now. Will have another PR in 15 minutes for the rest.

PR #338, opened 2019-10-31
tlindsay/security fixes

tlindsay:
Addresses security vulnerabilities in nested dependencies


snewcomer:
Awesome these vulnerabilities were resolved!  Tomorrow, let us see if we can track down the offending already installed pkgs and see if ppl have opened issues.  👍

PR #341, opened 2019-11-20
Updating intersect helper to work with ManyArrays

gmurphey:

Currently the `intersect` helper returns an empty array when using an Ember Data has many relationship. This PR updates the way intersections are tested and accesses the members of a given array in a consistent and predictable way.

Thanks for the great addon!


addon/helpers/intersect.js
```diff
@@ -9,9 +9,9 @@ export function intersect([...arrays]) {
   let results = confirmedArrays.pop().filter(candidate => {
     for (let i = 0; i < arrays.length; i++) {
       let found = false;
-      let array = arrays[i];
+      let array = emberArray(arrays[i]);
       for (let j = 0; j < array.length; j++) {
-        if (array[j] === candidate) {
+        if (array.objectAt(j) === candidate) {
```
snewcomer: One question @gmurphey before merging - what makes this code as is on master not able to work with Ember arrays?

snewcomer: ping!

PR #343, opened 2019-12-11
Restore helper functions exports

vladucu:
This restores some of the helper functions exports that were removed during #321


GavinJoyce:
Thanks for adding these back @vladucu 🍻.

@snewcomer perhaps we could merge this and cut a release?

PR #363, opened 2020-04-24
PoC mut-then and execute

betocantu93:
There are two helpers in this PR that cover two of our heavy use cases that we couldn't think of any combination of the current existing helpers could solve.

### mut-then Helper

In Ember `{{mut}}` helper is widely used, there's currently no helper to react to some mutation 

While `pipe` works great for processing functions and passing the return value to the next one, in some situations regarding mut, we don't care about about the actual changed value, but we do care about the "event" of something actually changing, like an observer

For example think about an autosave use case, we actually don't care at all about the return value of mut, which also is just undefined I think.
 
```hbs
<PaperInput @onChange={{pipe (fn (mut this.value)) (invoke "save" @model)}} />
```

While `queue` works great for passing the same argument to all the functions, again, in some use cases we just want to react. Using queue is worse because the onChange will be triggered with "my string" and "my string".save doesn't exist, so this is actually a bug.

```hbs
<PaperInput @onChange={{queue (fn (mut this.value)) (invoke "save" @model)}} />
```

I know we can hand write actions in our components to handle stuff like this, but the dev ergonomics and composability goes to the floor.

The mut-then (we can name it differently, honestly is just a PoC) proposal is the first function will receive arguments, and then keep executing the next ones without any args, so we can mutate and then harness the power of partial application without polluting the args.
 
```hbs
<PaperInput @onChange={{mut-then (fn (mut this.value)) (invoke "save" @model)}} />
```

```hbs
<PaperInput @onChange={{mut-then (fn (mut this.value)) (fn (pipe (invoke "save") onModelSaved) @model) }} />
```
And so on... 

### Execute Helper

Literally just execute all the actions without passing anything to them, let partial application do the magic, this is great for reacting to events like `onBlur`, for example we don't really wanna be saving `onChange` for an input, we probably wanna do it onBlur.

```hbs
<PaperInput @onBlur={{(execute (invoke "save" @model) onModelSaved}} />
```

### DIFF
--- a/addon/helpers/execute.js
+++ b/addon/helpers/execute.js
@@ -1,8 +1,8 @@
 import { helper } from '@ember/component/helper';
 import isPromise from '../utils/is-promise';
 
-export function mutThen(actions = []) {
-  return function(...args) {
+export function execute(actions = []) {
+  return function() {
     let invoke = function(acc, curr) {
       if (isPromise(acc)) {
         return acc.then(curr);
@@ -13,7 +13,7 @@ export function mutThen(actions = []) {
 
     return actions.reduce((acc, curr, idx) => {
       if (idx === 0) {
-        return curr(...args);
+        return curr();
       }
 
       return invoke(acc, curr);
@@ -21,4 +21,6 @@ export function mutThen(actions = []) {
   };
 }
 
-export default helper(mutThen);
+export default helper(execute);
+
+
--- a/addon/helpers/mut-then.js
+++ b/addon/helpers/mut-then.js
@@ -1,0 +1,24 @@
+import { helper } from '@ember/component/helper';
+import isPromise from '../utils/is-promise';
+
+export function mutThen(actions = []) {
+  return function(...args) {
+    let invoke = function(acc, curr) {
+      if (isPromise(acc)) {
+        return acc.then(() => curr());
+      }
+
+      return curr();
+    };
+
+    return actions.reduce((acc, curr, idx) => {
+      if (idx === 0) {
+        return curr(...args);
+      }
+
+      return invoke(acc, curr);
+    }, undefined);
+  };
+}
+
+export default helper(mutThen);
--- a/app/helpers/execute.js
+++ b/app/helpers/execute.js
@@ -1,0 +1,1 @@
+export { default, execute } from 'ember-composable-helpers/helpers/execute';
--- a/app/helpers/mut-then.js
+++ b/app/helpers/mut-then.js
@@ -1,0 +1,1 @@
+export { default, mutThen } from 'ember-composable-helpers/helpers/mut-then';
--- a/tests/integration/helpers/execute-test.js
+++ b/tests/integration/helpers/execute-test.js
@@ -1,0 +1,17 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Helper | execute', function(hooks) {
+  setupRenderingTest(hooks);
+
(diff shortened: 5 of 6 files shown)


addon/helpers/execute.js
```diff
@@ -0,0 +1,26 @@
...
+      }
+
+      return invoke(acc, curr);
+    }, undefined);
+  };
+}
+
+export default helper(execute);
+
+
```
snewcomer, line 26: So this is a copy of `queue` without args?  What is the pain of ignoring the args in each `action`?

betocantu93: Yes that's correct, a copy but without args, mainly to give the template ergonomics, Currently there's no way to simple chain events giving the template full control of partial application without leaks, queue and pipe 'leaks' into the chain of fns, for example using `on` with queue helper leaks the Dom event to the chain, using pipe leaks the return value, which might actually cause troubles if any fn in the chain returns void and any of the next function args expects maybe params  to be defined or something, you can see execute as the `call` helper but with a promise aware chain of fns

 To chain perhaps unrelated stuff like closing a dialog, creating a contact, displaying a notification and also triggering `@onClick` and logging to Google Analytics via a service action, all of these fns may receive args in certain way or order so leaking may have undesirable effects

```typescript
function something(str, isValid) {
  console.log(str)
  if(isValid){
     this.model.save()
  }
}
```

```handlebars
<button {{on "click" (queue this.openModal (fn this.something "hey there")}} >

Click me 

<button />

```

The event will be leaked, and so isValid would be true... I can try to give more examples if that doesn't make sense yet?
 
In our use case we have a serious amount of fns down on the render tree which need to hook say to the onBlur of an input and it wouldn't be posible if we had to worry about how they are actually physically placed, we see them as kind of a series of reactions to an event which of them is entirely responsible of partially apply the needed params

betocantu93: Maybe call helper may receive a chain? Would that make more sense? 🤔🤔

betocantu93: Yup, totally agree, what do you think about sequence?

- sequence

🤔🤔🤔

snewcomer: Also, it sounds like the functions you are using might be called in multiple places?  Otherwise, you could always align the arguments with the arity of arguments if only used in one spot?

I'm also thinking about this from a typescript point of view.  You have a list of expected arguments.  However, this will pass nothing and could show an error.  Typically, in this case you would compose multiple functions together while perhaps sharing some logic.

What are your thoughts?

betocantu93: > Also, it sounds like the functions you are using might be called in multiple places? Otherwise, you could always align the arguments with the arity of arguments if only used in one spot?

Yes, pretty much. I use them sparingly across the app/engines in different ways, from JS and helpers, so having full control on the template is a must, to avoid foot guns.

> I'm also thinking about this from a typescript point of view. You have a list of expected arguments. However, this will pass nothing and could show an error. Typically, in this case you would compose multiple functions together while perhaps sharing some logic.

Honestly I haven't (probably should) used a lot of functional programming or composition of fns, in `.js` land in Ember. I often use the templates for that, but I agree, you compose multiple functions together taking the arity, order and such into account when needed to share logic or something in common, in this particular case, the only thing these fns have in common is that they were triggered by something... sometimes an event, sometimes a callback, basically like observer pattern, pretty much, these are subscribers fns. in a Template only Component land, pipe, queue, (sequence|execute) come pretty handy

betocantu93: Another good approach could be to just add a `noargs` helper...

```hbs
<button {{on "click" (queue fn1 fn2 (noargs (fn fn3 myWantedArg)))}}>
```

```ts
function noargs(fn) {
  return () => fn()
}
```

PR #367, opened 2020-06-03
Fix sort-by to accept Ember Data relationship

jrjohnson:
The refactor of `sort-by` in #362 by @snewcomer no longer accepts a resolved Ember Data hasMany relationship as [this line](https://github.com/DockYard/ember-composable-helpers/blob/master/addon/helpers/sort-by.js#L28) throws `array is not iterable`.

### DIFF
--- a/addon/helpers/sort-by.js
+++ b/addon/helpers/sort-by.js
@@ -24,6 +24,9 @@ function sortAsc(key, a, b) {
 class SortBy {
   constructor(...args) {
     let [array] = args;
+    if (typeof array.toArray === "function") {
+      array = array.toArray();
+    }
 
     this.array = [...array];
     this.callbacks = null;
--- a/tests/integration/helpers/filter-test.js
+++ b/tests/integration/helpers/filter-test.js
@@ -89,4 +89,29 @@ module('Integration | Helper | {{filter}}', function(hooks) {
 
     assert.equal(find('*').textContent.trim(), 'this is all that will render', 'no error is thrown');
   });
+
+  test('it accepts a fulfilled ember data promise as a value', async function (assert) {
+    let store = this.owner.lookup('service:store');
+    let person = store.createRecord('person');
+
+    person.get('pets').pushObjects([
+      store.createRecord('pet', { name: 'aa' }),
+      store.createRecord('pet', { name: 'ab' }),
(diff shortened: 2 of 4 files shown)


snewcomer:
Thanks for the PR!  This seems like it would be a problem for all the helpers, not just sort-by.  I'm not exactly sure of a solution off the top of my head but for awaiting the promise before passing into any of the ember-composable-helpers...

jrjohnson:
I'm not sure either, it works everywhere else at this point (and worked in `sort-by` in 4.0.0). I didn't want to jam up this PR with more tests, but I'll add those tests for all the remaining helpers once this is resolved here to prevent another regression. 

Our particular use case is to pass an ember data relationship in using the [await](https://github.com/fivetanley/ember-promise-helpers#await) helper like
```
{{#each (sort-by "lastName" @course.directors) as |user|}}
{{/each}}
```

So the promise is actually resolved, but it's passing in a resolved promise and not an array. I couldn't come up with an elegant way to check if something was an already resolved promise and pull the value, but that is probably what is needed (as the helpers, in my opinion, shouldn't be async).

snewcomer:
Feel free to add some more tests to this PR!  I was looking at filter and was curious how that would work wrapped in a Promise as well.

jrjohnson:
You're right! Sorry for the red-herring. The issue isn't the promise, but the `EmberArray` that is resolved from a hasMany Ember Data relationship. Which points to a fairly simple fix. I also added a couple tests in other places while I was experimenting and left those in for posterity.

GavinJoyce:
👍 nice one @jrjohnson, I tried to upgrade `ember-composable-helpers` to the latest today and had a bunch of failing tests. This PR fixes those tests

mansona:
I've just come across this PR now too. I'm pretty sure https://github.com/DockYard/ember-composable-helpers/pull/362 is a breaking change for anything ember-data related and not just relationships. The case that broke for me was just a simple: 

`{{#each (sort-by 'date:desc' this.model) as |post index|}}`

so I'm not actually accessing anything via a relationship and I have the same regression. 

@snewcomer if we aren't going ahead with this PR can we at least revert #362 and get a patch release out?

jrjohnson:
The promise thing was me over-simplifying the problem (incorrectly) and not actually something we need. This fix resolves all of our issues.

PR #368, opened 2020-06-04
add a changelog

mansona:
Hey folks 👋  this is mostly to open a discussion about adding a changelog. 

I created one really quick using https://www.npmjs.com/package/auto-changelog so I've not spent any time on this 😂 

The reason I'm bringing up this discussion is that there seems to be some regression between 3.x and 4.x of this addon and I wanted to investigate what changed, but there was no documentation in a changelog or the releases.


mansona:
Turns out that having a changelog really helped me identify the problem 🎉  it is actually a regression between 4.0 and 4.1 because of https://github.com/DockYard/ember-composable-helpers/pull/362

I've commented on the PR that looks to fix the issue to mention that I'm seeing the same problem 👍

jrjohnson:
Either a change log or github release notes for each version would make upgrading much easier. I tend to prefer the release notes personally as they are easier to build into a deploy process since the CHANGELOG has to be committed before a release and tends to be a partially manual process.

mansona:
@jrjohnson apparently there is a trick with this auto-changelog that means it will automatically update your changelog when you run `npm version` or `yarn version` to update the package.json version and tag the repo. It will also automatically work with `np` so if that is already part of the release process this can be 100% automated 😱

PR #376, opened 2020-06-12
Further improve sorting algorithm to put undefined values at end of list

snewcomer:
close #374

### DIFF
--- a/addon/helpers/sort-by.js
+++ b/addon/helpers/sort-by.js
@@ -7,7 +7,11 @@ function normalizeToBoolean(val) {
   }
 
   if (typeof val === 'number') {
-    return val > 0 ? false : true;
+    if (val > 0) {
+      return false;
+    } else if (val < 0) {
+      return true;
+    }
   }
 
   return val;
@@ -16,6 +20,15 @@ function normalizeToBoolean(val) {
 function sortDesc(key, a, b) {
   const aValue = get(a, key);
   const bValue = get(b, key);
+
+  if (typeof bValue == 'undefined') {
+    // keep bValue last
+    return -1;
+  }
+  if (typeof aValue == 'undefined') {
+    // put aValue last
+    return 1;
+  }
 
   if (aValue && bValue && aValue.toLowerCase && bValue.toLowerCase) {
     return bValue.localeCompare(aValue, undefined, { sensitivity: 'base' });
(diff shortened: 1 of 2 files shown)


tests/integration/helpers/sort-by-test.js
```diff
@@ -44,7 +44,7 @@ module('Integration | Helper | {{sort-by}}', function(hooks) {
       {{~/each~}}
     `);
 
-    assert.equal(find('*').textContent.trim(), 'aAAabccb', 'sorts multiletter words');
+    assert.equal(find('*').textContent.trim(), 'AaaAbccB', 'sorts multiletter words');
```
snewcomer, line 47: We aren't splitting the strings up by individual char, so I actually think this is now "more" right.

addon/helpers/sort-by.js
```diff
@@ -7,7 +7,11 @@ function normalizeToBoolean(val) {
   }
 
   if (typeof val === 'number') {
-    return val > 0 ? false : true;
+    if (val > 0) {
+      return false;
+    } else if (val < 0) {
+      return true;
+    }
```
snewcomer, line 14: We shouldn't swap if `val === 0`, thus the need for this refactor

snewcomer:
@jrjohnson Can I get your careful eye on this PR?  Thank you!!

addon/helpers/sort-by.js
jrjohnson: Ah! Yes! This is what I could NOT seem to get my head around and fix. Yay!

addon/helpers/sort-by.js
```diff
@@ -17,6 +21,15 @@ function sortDesc(key, a, b) {
   const aValue = get(a, key);
   const bValue = get(b, key);
 
+  if (typeof bValue == 'undefined') {
```
jrjohnson, line 24: This makes more sense to me than my fix.

addon/helpers/sort-by.js
```diff
@@ -17,6 +21,15 @@ function sortDesc(key, a, b) {
...
+    // keep bValue last
+    return -1;
+  }
+  if (typeof aValue == 'undefined') {
+    // put aValue last
+    return 1;
+  }
+
   if (aValue && bValue && aValue.toLowerCase && bValue.toLowerCase) {
```
jrjohnson, line 33: `aValue && bValue && ` can now be removed here as it was just to catch `undefined`. I think everything else can have `toLowerCase` check called on it.

addon/helpers/sort-by.js
```diff
@@ -34,6 +47,15 @@ function sortAsc(key, a, b) {
...
```
jrjohnson, line 59: Same,  `aValue && bValue && ` can go away now that undfined check is above.

snewcomer:
Thank you for the awesome review!!  4.2.1 released!!

PR #379, opened 2020-08-25
Rename {{contains}} helper to {{includes}}

### LINKED ISSUE #365: Rename contains helper to includes
The naming of `{{contains}}` helper is not intuitive as it does not map the name of the native JavaScript method: `Array.includes()`. It should be renamed to `{{includes}}`. The old name should be deprecated.

Naming it *includes* instead of *contains* [was already considered before implementation](https://github.com/DockYard/ember-composable-helpers/issues/13#issuecomment-186334552) back in 2016 for the same reason. It's not documented why Lauren decided to go with *contains*. It might have been named after `contains` method of `Ember.Array` and `Ember.Enumerable`. That ones [have been deprecated](https://deprecations.emberjs.com/v2.x/#toc_ember-runtime-enumerable-contains) and removed in 2.x.

jelhan:
This renamed `{{contains}}` helper to `{{includes}}` as discussed in #365. A deprecation warning is added for the old name.

Additional I have refactored to code:

- Remove code only needed to support Ember versions, which fall back to `EmberArray.contains()` if `EmberArray.includes()` is not available. `EmberArray.contains()` has been removed in 2.x. `EmberArray.includes()` is available in all supported version.
- Avoid converting the array passed in multiple times to `EmberArray` if an array of needles is provided.
- Use `Array.every()` (or `EmberArray.every()`) instead of `Array.reduce()` to improve readability and performance.
- Use same code path for single needle and array of needles to reduce complexity. This may have a bad performance impact as it requires creating an interim array if a single needle is passed. But the performance impact should be small enough to be acceptable.

Closes #365

### DIFF
--- a/addon/helpers/contains.js
+++ b/addon/helpers/contains.js
@@ -1,18 +1,18 @@
-import { A as emberArray } from '@ember/array';
-import { isArray as isEmberArray } from '@ember/array';
 import { helper } from '@ember/component/helper';
+import { includes } from './includes';
+import { deprecate } from '@ember/debug';
 
 export function contains(needleOrNeedles, haystack) {
-  if (!isEmberArray(haystack)) {
-    return false;
-  }
+  deprecate(
+    '{{contains}} helper provided by ember-composable-helpers has been renamed to {{includes}}.',
+    false,
+    {
+      id: 'ember-composable-helpers.contains-helper',
+      until: '5.0.0'
+    }
+  );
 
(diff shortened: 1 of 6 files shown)


Alonski:
@jelhan @snewcomer Not to be picky but the README wasn't updated with this change 😜

bertdeblock:
Created https://github.com/DockYard/ember-composable-helpers/pull/397 for the docs part.

PR #380, opened 2020-09-02
Add call helper

### LINKED ISSUE #264: Optional helper with noop instead of identity result?
I hit this use case in our app, but the usage of actions in this particular case is not exactly orthodox, because we call the function directly and use its result, instead of using `sendAction`.

I solved this issue by creating a custom helper, similar to the `optional` one included in ember-composable-helpers, but I was wondering if it would make sense as an addition to the library.

I can submit a PR, but I wanted to confirm the usefulness of this idea.

snewcomer:
close #264

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ Watch a free video overview presented by EmberMap:
       - [`compact`](#compact)
       - [`contains`](#contains)
       - [`append`](#append)
+      - [`call`](#call)
       - [`chunk`](#chunk)
       - [`without`](#without)
       - [`shuffle`](#shuffle)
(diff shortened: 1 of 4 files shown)


MrChocolatine:
I am a bit confused, what is the difference with the `{{compute}}` helper? They both seem to do the same thing, take a function, with or without arguments, execute + return it.

snewcomer:
`compute` doesn't keep context I believe.

```
{{compute (action this.foobar)}}
```

MrChocolatine:
You are right, the only subtle difference in `{{call}}` would be the extra context we can pass. To me it seems we can (should?) merge these two helpers as they seem way too similar in what their purpose is.

https://github.com/DockYard/ember-composable-helpers/blob/5754b4195581394efa261ac22ad3d269d70f62f5/addon/helpers/compute.js

https://github.com/DockYard/ember-composable-helpers/blob/5754b4195581394efa261ac22ad3d269d70f62f5/addon/helpers/call.js

Would you accept someone work on that idea or do you prefer to keep things like this?

Issue #385, opened 2020-11-09
Sort order when no key provided has changed between 4.1.1 and 4.3.0


GavinJoyce:
There seems to be a difference in sort order between `4.1.1` and `4.3.0`.

```js
array = [{ name: 'c' }, { name: 'a' }, { name: 'b' }];
```

```hbs
{{~#each (sort-by "" this.array) as |user|~}}
  {{~user.name~}}
{{~/each~}}
```

In `4.1.1`, this renders unsorted as `cab`.
In `4.3.0` it renders as `bac`.

GavinJoyce:
Fix here: https://github.com/DockYard/ember-composable-helpers/pull/386

GavinJoyce:
fix here: https://github.com/DockYard/ember-composable-helpers/pull/386

PR #386, opened 2020-11-09
`{{sory-by}}` - Don't sort array when key is empty

GavinJoyce:
Fixes https://github.com/DockYard/ember-composable-helpers/issues/385

/cc @snewcomer


snewcomer (approved):
👍 👍

PR #388, opened 2020-11-18
[sort-by] Cannot call get with [keyName] on an undefined object.

### LINKED ISSUE #387: [sort-by]  Cannot call get with 'configOrder' on an undefined object.
looks like it may be an regression after https://github.com/DockYard/ember-composable-helpers/pull/362/files

related issue: https://github.com/DockYard/ember-composable-helpers/issues/369

## Version
4.3.1

working `ok` on: `3.2.0`, `4.0.0`
breaking from: `4.1.0`

## Steps to reproduce
```js
let items = [{configOrder:1}, undefined]
```

```hbs
{{sort-by 'configOrder' items}}
```

## Expected Behavior
sort should work, including "holes" in array

## Actual Behavior
error:  Cannot call get with [sortKeyName] on an undefined object.

## Why? 
because Ember.get don't accept undefined, null as first argument and throw error

## Error example:
https://ember-twiddle.com/0e2d286133936b27d0de191dac08ddfb?openFiles=templates.application%5C.hbs%2C (check console)

lifeart:
Closes #387

added check for null or undefined, before ember's get usage to prevent mentioned errors

### DIFF
--- a/addon/helpers/sort-by.js
+++ b/addon/helpers/sort-by.js
@@ -18,13 +18,20 @@ function normalizeToBoolean(val) {
   return val;
 }
 
+function safeValueForKey(ctx, key) {
+  if (ctx === null || ctx === undefined) {
+    return ctx;
+  }
+  return get(ctx, key);
+}
+
 function sortDesc(key, a, b) {
   if (isEmpty(key)) {
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


snewcomer (approved):
Phenomenal!

Issue #364, opened 2020-04-25
Using undefined arrays with array helpers results in exception @3.x and @4.x, but was working in 2.4


roberkules:
## Version
3.x and 4.x

## Test Case
version | status | ember-twiddle
------- | ------ | -------------
2.4.0 | working | https://ember-twiddle.com/50396ad6e489d231abbbd18c4877fc73?openFiles=templates.application%5C.hbs%2C
4.0.0 | failing | https://ember-twiddle.com/9aa2bc40db05617dd5133be13f00da22?openFiles=templates.application%5C.hbs%2C

## Steps to reproduce
use a array helper (I used `map`) with an undefined array

## Expected Behavior
array helpers invoked with undefined array should return empty array, as in version 2.x (or it should be marked as breaking change)

## Actual Behavior
`Uncaught TypeError: Cannot read property 'map' of undefined`

---

before `v3.x` ember-composable-helpers used computed macros:

### 2.4.0

```javascript
import { map } from '@ember/object/computed';
// ...
defineProperty(this, 'content', map('array', callback));
```

### 3.x / 4.x

```javascript
return array.map(callback);
```

snewcomer:
@roberkules We would gladly accept a PR!

Issue #269, opened 2017-05-13
Frozen array causes errors with find-by


ming-codes:
`find-by` and possibly other helpers that wraps using `Ember.A` throws error.

## Version
1.1.2, 2.0.1

## Test Case
https://ember-twiddle.com/e5184a790c517817d6a64650ddb88ea7?openFiles=templates.application.hbs%2C

## Steps to reproduce
https://ember-twiddle.com/e5184a790c517817d6a64650ddb88ea7?openFiles=templates.application.hbs%2C

## Expected Behavior
No error!

## Actual Behavior
Error thrown

anehx:
Happens with `{{includes}}` as well, see https://ember-twiddle.com/94cbfbbea6e38e60d1966b19fbd9dc60

PR #389, opened 2020-11-21
better array handlers (for nan, null, undefined, sets)

lifeart:
fixes: https://github.com/DockYard/ember-composable-helpers/issues/364
fixes: https://github.com/DockYard/ember-composable-helpers/issues/269

Adding new `asArray` function, wich is trying to return array object, for any actions on it.

Throw error on:

1. Numbers
1. Symbol
1. EmberObject without toArray property
1. WeakMap
1. WeakSet
1. Non-iterables
1. Promise-like object, without `content` property
1. Promise-like object, with `content` property, equals to null

Now working with:

1. Set
1. Map
1. String (string.split(''))
1. Plain Object (Object.values() used as array content)
1. null (fallback to empty array)
1. undefined (fallback to empty array)
1. Promise-like object, with `content` property (using content as possible array, with asArray transform)

### DIFF
--- a/addon/helpers/includes.js
+++ b/addon/helpers/includes.js
@@ -1,6 +1,7 @@
 import { A as emberArray } from '@ember/array';
 import { isArray as isEmberArray } from '@ember/array';
 import { helper } from '@ember/component/helper';
+import asArray from '../utils/as-array';
 
 export function includes(needleOrNeedles, haystack) {
   if (!isEmberArray(haystack)) {
@@ -10,7 +11,7 @@ export function includes(needleOrNeedles, haystack) {
   let needles = isEmberArray(needleOrNeedles) ? needleOrNeedles : [needleOrNeedles];
   let haystackAsEmberArray = emberArray(haystack);
 
-  return needles.every((needle) => {
+  return asArray(needles).every((needle) => {
     return haystackAsEmberArray.includes(needle);
   });
 }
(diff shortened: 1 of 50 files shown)


lifeart:
thing to disuss:
`asArray` added supports for stings `asArray("abc") -> ["a","b","c"]`, wondering, should we support it? same for sets, maps, and objects itself (object.values)

lifeart:
also, should this this issue: 
image

snewcomer:
I like this PR!  Happy to merge when ready!

addon/helpers/includes.js
```diff
@@ -10,7 +11,7 @@ export function includes(needleOrNeedles, haystack) {
   let needles = isEmberArray(needleOrNeedles) ? needleOrNeedles : [needleOrNeedles];
   let haystackAsEmberArray = emberArray(haystack);
```
anehx, line 12:
```suggestion
  let haystackAsEmberArray = emberArray(asArray(haystack));
```
This would possibly fix the issue with frozen arrays and the `{{includes}}` helper. I created a fiddle for reproduction: https://ember-twiddle.com/94cbfbbea6e38e60d1966b19fbd9dc60

lifeart:
@snewcomer I think I'm done with all planned changes

snewcomer (approved):
This is a phenomenal addition to this addon.  Thank you. 🙇‍♂️ 

I'll do a minor bump soon!  Again, thank you!!

PR #390, opened 2020-12-03
fixes sort-by case with multiple properties and nulls

yrakova:
Fixed case where some sorting values are null and other sorting keys doesn't applied


snewcomer (approved):
Thank you!!

PR #398, opened 2021-02-16
Remove tryInvoke (from @ember/utils) calls

brunoocasali:
- Remove `tryInvoke` calls from `invoke` helper and replace with javascript optional chaining.

Related to: https://emberjs.github.io/rfcs/0673-deprecate-tryinvoke.html

### DIFF
--- a/addon/helpers/invoke.js
+++ b/addon/helpers/invoke.js
@@ -1,6 +1,5 @@
 import { isArray as isEmberArray } from '@ember/array';
 import { helper } from '@ember/component/helper';
-import { tryInvoke } from '@ember/utils';
 import RSVP from 'rsvp';
 
 const { all } = RSVP;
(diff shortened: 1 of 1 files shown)


addon/helpers/invoke.js
```diff
@@ -10,14 +9,14 @@ export function invoke([methodName, ...args]) {
 
   if (isEmberArray(obj)) {
     return function() {
-      let promises = obj.map((item) => tryInvoke(item, methodName, args));
+      let promises = obj.map((item) => item[methodName]?.(args));
```
bertdeblock, line 12:
```suggestion
      let promises = obj.map((item) => item[methodName]?.(...args));
```

addon/helpers/invoke.js
```diff
@@ -10,14 +9,14 @@ export function invoke([methodName, ...args]) {
...
 
       return all(promises);
     };
   }
 
   return function() {
-    return tryInvoke(obj, methodName, args);
+    return obj[methodName]?.(args);
```
bertdeblock, line 19:
```suggestion
    return obj[methodName]?.(...args);
```

bertdeblock (commented):
`tryInvoke`'s last argument should be an array, but with the native approach we need to spread the args to achieve the same result. Atm the tests still pass by accident, but maybe this should also be tested more explicitly?

PR #403, opened 2021-04-02
Embroider Improvements

alexlafroscia:

I've been investigating Embroider in my app and have noticed that `ember-composable-helpers` generates some warnings about invalid exports. I figured it would be good to fix those!

While here, I noticed that the repo is configured to test against Embroider support with the style popularized easier in the Embroider development process. Now there's a first-party addon testing approach, which seems like the better way to ensure compatibility!

### DIFF
--- a/app/helpers/values.js
+++ b/app/helpers/values.js
@@ -1,1 +1,1 @@
-export { default, values } from 'ember-composable-helpers/helpers/values';
+export { default } from 'ember-composable-helpers/helpers/values';
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,8 @@ jobs:
             ember-beta,
             ember-canary,
             ember-default-with-jquery,
-            embroider-tests,
+            embroider-safe,
+            embroider-optimized
           ]
     continue-on-error: ${{ matrix.try-scenario == 'ember-beta' || matrix.try-scenario == 'ember-canary' }}
     steps:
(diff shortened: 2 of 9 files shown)


snewcomer:
How can I help here?  Are there still some unanswered questions?

alexlafroscia:
Great question. I think it depends on how much you all care about getting CI working again. 

If you guys want to make Beta and Canary tests optional then merging #405 first probably makes sense. 

If you don't want to make those tests optional, I'll close that PR and can continue hacking on #404 until I can figure out why it doesn't seem to be working in CI, even though it works locally. 

This can also just be merged right away if it's OK that CI starts failing in the main branch.

alexlafroscia:
Landing #405 first probably makes the most sense, since there are some good clean-up efforts in there around how CI runs, in particular the shorter timeout which makes iterating on getting CI to work much faster (due to the much shorter build timeout threshold). As a non-owner of this package I can't kill the jobs myself once it's clear it's going to hang forever.

app/helpers/values.js
```diff
@@ -1 +1 @@
-export { default, values } from 'ember-composable-helpers/helpers/values';
+export { default } from 'ember-composable-helpers/helpers/values';
```
alias-mac, line 1: I took a different approach in #406 here since other helpers are providing re-exports too.

alexlafroscia:
Sorry I never got this over the finish-line myself! I'm happy to see @alias-mac land #406 -- I'm going to close this.

PR #404, opened 2021-04-03
Fix Beta/Canary Tests

alexlafroscia:

The tests for #403 show the `ember-beta` and `ember-canary` builds timing out. This should address those problems and get them working again!

It also fixes some warnings related to Ember 4.0 that are printed when running the tests.

A few code mods were used to migrate to better-recommended approaches to things:

- [`ember-cli-htmlbars-inline-precompile-codemod`](https://github.com/ember-codemods/ember-cli-htmlbars-inline-precompile-codemod)
- [`qunit-dom-codemod`](https://github.com/simplabs/qunit-dom-codemod)

The `qunit-dom` change was made to resolve an issue where `find('*')` was used to check the text content of the whole "template" for an integration test, but that syntax isn't actually valid. I thought that `qunit-dom` might help. That wasn't _quite_ the answer, as I still had to go remove the `'*'` argument that was code-modded into the `qunit-dom` syntax, but the `qunit-dom` error made it more clear what was going wrong. Since the syntax is cleaner and the tests read more easily with them using that library, I figured I would leave that in this PR.


alexlafroscia:
I think the _actual_ culprit here was the `ember-cli-babel` version; it was odd that in CI, nothing was bring printed for the `test` output in the Beta and Canary builds. I think the problem was too many warnings being printed to the browser, stemming from the outdated `ember-cli-babel` version transpiling code to use the Ember global, which is now deprecated in those Ember versions.

alexlafroscia:
Okay, so the `ember-qunit` and `ember-resolver` respectively resolve some of the deprecation warnings about the `Ember` global.

To get the rest of them, we need to upgrade more dependencies that depend themselves on `ember-cli-babel`; that's the "root cause" of some of these issues.

alexlafroscia:
^ I'm hoping that cleaning up more of the deprecation warnings will show us whatever is causing the `embroider` test case to not show any kind of results log, now that the beta and canary tests are passing!

alexlafroscia:
With those last dependency updates, I was able to get the Embroider build to pass locally... 🤞 that it passes in CI, too!

alexlafroscia:
Same issue as before is still happening in CI, but Embroider installed locally seems to work fine... I don't get it 🤷

alexlafroscia:
Still working on reproducing that specific `define`-related bug locally...

If I just run the normal Embroider build locally (by follow it's normal installation steps) it works fine.

If I try to run it through `ember-try` locally, though, I _always_ get hung up on an error of this flavor, though which dependency it complains about changes

CleanShot 2021-04-04 at 09 55 11@2x

alexlafroscia:
Welp, I thought that a hail-Mary attempt at regenerating the whole lockfile might fix things by upgrading some deep-down version of a package that might be causing the problems, but... Apparently not!

alexlafroscia:
I'm going to try the Embroider setup from #403 just to see if that, for some reason, helps...

alexlafroscia:
Nope... Same problem 😡

alexlafroscia:
Okay, I'm going to back off and take a different approach here. I think there are some CI configuration improvements that could be made, including allowing the Beta and Canary builds to fail (at least for now, where it seems like it's something on Ember's end rather than on this package's end).

I'm going to open a separate PR to address that clean-up, then revisit after that.

I tried _again_ to get the Embroider tests to fail locally but... they just don't. That's making me think it might have something to do with the CI environment itself. Nothing _seems_ strange but here are some outdated packages (like `ember-cli-eslint`) that could be removed to get the configuration more inline with the modern standard.

snewcomer:
Is this current with the main branch?  Should we update/merge with the ember-try failures as welll?

PR #405, opened 2021-04-05
CI Configuration Improvements

alexlafroscia:

While working on #404 I noticed a number of oddities in the GitHub Actions config. Linting being run multiple times, the default `ember test` command never being run, stuff like that.

This PR aims to straighten some of this out.

The major change is that `ember-beta` and `ember-canary` are **now allowed to fail**, which might be a risk, but can also be a big help when the actual problems are up-stream.


alexlafroscia:
I don't quite understand why the `continue-on-error` settings didn't make it so that the job passed even though those jobs failed...

alexlafroscia:
What's weird is that viewing the workflow from the Actions section of the site shows that the workflows passed, even though the PR status seems to indicate failure. Not sure what the deal there is...

https://github.com/DockYard/ember-composable-helpers/actions/runs/719632704

alexlafroscia:
My hunch is that the `continue-on-error` setting will be respected once this lands into the main branch but is ignored while in a PR; some of the GitHub Actions configuration options work that way from my experience. I'm not quite sure though 🤔 Might be worth merging just to find out.

snewcomer (approved):
I'm ok with that!  Thanks a ton.  Looking forward to resolving the other PRs you have open as well.

PR #406, opened 2021-04-16
Embroider

alias-mac:
### Use @embroider/test-setup
Ember addons benefit on using this addon provided by embroider directly.
Read more information about it [here](https://github.com/embroider-build/embroider/tree/master/packages/test-setup).

### Fix warnings provided by embroider build:
While running `ember try:one embroider-safe` or `ember try:one embroider-optimized` you will see warnings during the build:
```  
    WARNING in /private/var/folders/1m/cqqx124s5gj0dn95jmbp91p0000mgw/T/embroider/a0a719/node_modules/ember-data/node_modules/ember-inflector/index.js 4:0-48
    "export 'defaultRules' was not found in './lib/system'
     @ ./assets/dummy.js
    
    WARNING in ./helpers/entries.js 1:0-60
    "export 'entries' was not found in '../../../helpers/entries'
     @ ./assets/dummy.js
    
    WARNING in ./helpers/from-entries.js 1:0-69
    "export 'fromEntries' was not found in '../../../helpers/from-entries'
     @ ./assets/dummy.js
    
    WARNING in ./helpers/pick.js 1:0-54
    "export 'pick' was not found in '../../../helpers/pick'
     @ ./assets/dummy.js
    
    WARNING in ./helpers/values.js 1:0-58
    "export 'values' was not found in '../../../helpers/values'
     @ ./assets/dummy.js
```
The ones related with this addon are fixed (in the [second commit](https://github.com/DockYard/ember-composable-helpers/commit/cc62c932fb72105a7c9297e08898caed43676fba)).

I had to rename `keys` to `entries` to match the expectation in `app/` re-export, but this **can be considered a breaking change**, so perhaps it will be better to change the re-export to be `keys` instead of `entries`, although that breaks the convention from the other re-exports. Open to suggestions on how you want to proceed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,8 @@ jobs:
             ember-beta,
             ember-canary,
             ember-default-with-jquery,
-            embroider-tests,
+            embroider-safe,
+            embroider-optimized,
           ]
     continue-on-error: ${{ matrix.try-scenario == 'ember-beta' || matrix.try-scenario == 'ember-canary' }}
     steps:
--- a/addon/helpers/entries.js
+++ b/addon/helpers/entries.js
@@ -1,10 +1,10 @@
 import { helper } from '@ember/component/helper';
 
-export function keys([object]) {
+export function entries([object]) {
   if (!object) {
     return object;
   }
   return Object.entries(object);
 }
 
-export default helper(keys);
+export default helper(entries);
--- a/addon/helpers/from-entries.js
+++ b/addon/helpers/from-entries.js
@@ -1,8 +1,10 @@
 import { helper } from '@ember/component/helper';
 
-export default helper(function fromEntries([entries]) {
+export function fromEntries([entries]) {
   if (!entries) {
     return entries;
   }
   return Object.fromEntries(entries);
-});
+}
+
+export default helper(fromEntries);
--- a/addon/helpers/pick.js
+++ b/addon/helpers/pick.js
@@ -1,7 +1,7 @@
 import { helper } from '@ember/component/helper';
 import { get } from '@ember/object';
 
-export default helper(function event([path, action]/*, hash*/) {
+export function pick([path, action]/*, hash*/) {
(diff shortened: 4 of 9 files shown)


alias-mac:
This is related with #403 but I took a different approach (since I'm not removing the re-exports).

snewcomer:
@alias-mac Thanks for this PR!  I'm fine either way.  Would you mind merging in the main branch and see what happens to the test suite?

https://github.com/DockYard/ember-composable-helpers/pull/404

alias-mac:
@snewcomer rebased and ran the tests locally and they pass, but they are failing in CI.

I was able to reproduce it when I run it with `CI=true` environment too locally. After a bit of experimentation I was able to confirm that it is due to the fact that we test against [IE11 in CI](https://github.com/DockYard/ember-composable-helpers/blob/b253d2ebb961a5adf57b91cc8a3e7212648aa55e/tests/dummy/config/targets.js#L13). When I remove that it passes.

Based on [embroider's changelog](https://github.com/embroider-build/embroider/blob/master/CHANGELOG.md) it seems that IE11 is supported. I'll try to take a stab at it and see what might be needed to get this fixed, but open to ideas.

alias-mac:
@snewcomer I synced with @rwjblue and he pointed me in the right direction on what was needed to make this pass.
The [addon blueprint will contain new options by default](https://github.com/ember-cli/ember-new-output/blob/2ff8bf5b89e2997bce2370fa012df804bed18a06/ember-cli-build.js#L25-L29), so I just added them here.
The [test-setup function](https://github.com/embroider-build/embroider/blob/8b2c20d6dccb006c4cc51cb18504f5a489c948f2/packages/test-setup/src/index.ts#L16) should probably support that by default, but for now this solves the problem.

Let me know what are the next steps to get this in. Thanks!

snewcomer (approved):
Woop woop!

rwjblue:
@snewcomer any chance for a release with this one?

PR #422, opened 2021-12-06
Switch from yarn to npm

cibernox:


snewcomer (approved):

.github/workflows/ci.yml
```diff
@@ -18,9 +18,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
-          cache: yarn
```
snewcomer, line 21: I think we can keep this around with `cache: npm` if you so desire!
 
https://github.com/actions/setup-node

.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
...
```
snewcomer, line 65: I think we could prob ✂️  `ember-default-with-jquery`

snewcomer (approved):
woo hooo!

cibernox:
Finally! Now I'll release 5.0 if I have permissions

ctjhoa:
@cibernox Can you provide more info about that change? I'm looking at porting this addon to v2 format however yarn seems to be the recommended solution
https://github.com/embroider-build/embroider/blob/main/PORTING-ADDONS-TO-V2.md

snewcomer:
Both are fine.  So is pnpm!  Yarn is used in the example because we haven't switched to another pkg tool!

PR #435, opened 2022-10-10
[maybe breaking] Stop using Ember.A()

jrjohnson:

Ember's Array prototype extensions are being [deprecated soon](https://rfcs.emberjs.com/id/0848-deprecate-array-prototype-extensions), and Ember Data has [already deprecated them](https://rfcs.emberjs.com/id/0745-ember-data-deprecate-methods-on-promise-many-array). As part of Ember Data's deprecation work calling Ember.A on a PromiseMany [has been deprecated](https://github.com/emberjs/data/blob/47a71ca1538ba9e2d7dfa01bf048a2db897bdf5f/packages/store/addon/-private/record-arrays/identifier-array.ts#L385) so removing any calls to this on passed values in composable helpers is needed.

For the most part the usage of Ember.A here were replaced with native Array.includes which is supported everywhere but IE 11. This dropping of IE 11 support is why I marked this PR as breaking even though nothing in the API surface changed.

Unfortunately Array.at() does not yet have universal support so I've added a polyfill for that copied from the TC39 proposal.

### DIFF
--- a/addon/helpers/find-by.js
+++ b/addon/helpers/find-by.js
@@ -1,6 +1,6 @@
 import { helper } from '@ember/component/helper';
 import { isEmpty } from '@ember/utils';
-import { A as emberArray } from '@ember/array';
+import { get } from '@ember/object';
 import asArray from '../utils/as-array';
 
 export function findBy([byPath, value, array]) {
@@ -8,7 +8,9 @@ export function findBy([byPath, value, array]) {
     return [];
   }
 
-  return emberArray(asArray(array)).findBy(byPath, value);
+  return asArray(array).find((item) => {
+    return get(item, byPath) === value;
+  });
 }
 
 export default helper(findBy);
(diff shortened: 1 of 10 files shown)


sly7-7:
@jrjohnson Do you think you could also remove [this toArray call](https://github.com/DockYard/ember-composable-helpers/blob/43aeb7f31c2d3ebfb138df2e4468132e656d08f8/addon/helpers/sort-by.js#L112) call, since ember-data deprecates its usage ? ```The `toArray` method on the class ManyArray is deprecated.```
🤔 Maybe this has to be an other PR

jrjohnson:
@sly7-7 I think that probably needs it's own PR because there are a bunch of calls to `toArray` scattered around and I'm not sure which ones can just be `slice`. Some of it may conflict with this change though, so this probably needs to get merged first. Once that is done though I'm happy to take a pass at removing `toArray` everywhere and it can go into the same breaking release as this.

sly7-7:
@jrjohnson This makes sense, thanks a lot.

jrjohnson:
@snewcomer I see you did the last release here, are you the right person to ping to get this merged?

PR #453, opened 2024-05-15
⚠️ add declaration of non maintenance, and what to do about it

NullVoxPopuli:
This library has been forked, and improved.

The original copy, while historically influential and very useful, should not be used in new work. GJS and GTS components allow folks to just use javascript, and not need helper work-arounds.

This PR adds to the readme describing what to do.

This repo may also want to consider archival?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,20 @@
 # ⚠️ Unmaintained ⚠️ Do not use ⚠️
 
-Use this fork instead:
+With GJS / GTS / `<template>` components, most of composable-helpers is no longer needed.
+Just use JavaScript.
+For example:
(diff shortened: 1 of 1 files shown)


gorner:
Link to OP's fork for the lazy: https://github.com/nullvoxpopuli/ember-composable-helpers

Though I see there has been at least one other recent attempt at a modernized fork: https://github.com/eflexsystems/ember-composable-helpers

NullVoxPopuli:
Mine is a v2 addon and I removed all the build complexity, if that matters to folks.
I also added some stats on build size -- the whole library is 13k, post min+brotli

